PR #9, opened 2019-04-23
Check for type of MessageEvent source before closing

chrisscott:
Make sure it is a Window before attempting to `close()` it.

Prevents a fatal error when using `getTokenSilently()`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -30,7 +30,7 @@ export const runIframe = (authorizeUrl: string, eventOrigin: string) => {
     const iframeEventHandler = function(e: MessageEvent) {
       if (e.origin != eventOrigin) return;
       if (!e.data || e.data.type !== 'authorization_response') return;
-      (<any>e.source).close();
+      if (e.source instanceof Window) (<any>e.source).close();
       e.data.response.error ? rej(e.data.response) : res(e.data.response);
       clearTimeout(timeoutSetTimeoutId);
       window.removeEventListener('message', iframeEventHandler, false);


chrisscott:
I'm not sure the best way to resolve the tests failing since in practice `e.source.close()` doesn't always exist as `e.source` isn't always a Window object. Simplest seems to remove the tests since we can't guarantee `close()` will always run but I wanted to get feedback before removing them.

luisrudge:
In what case the instance will not be a Window?

chrisscott:
In the browser, I'm seeing it also be `undefined` or `global`.

luisrudge:
@chrisscott can you tell me what are the repro steps for being undefined? Also, what is `global` in this context?

chrisscott:
@luisrudge  I'm seeing `iframeEventHandler()` being called twice when calling `createAuth0Client()`. The first time the former it is called, `e.source` is a Window but the second time it shows as `global` in the debugger. Setting a breakpoint in `iframeEventHandler()` should hopefully show what I'm seeing. 

Semi-related, should the negation in [this conditional check](https://github.com/auth0/auth0-spa-js/blob/master/src/index.ts#L10) not be there? Seems like the intent is to run `getTokenSilently()` when  the user isn't authenticated but this is causing the opposite.

luisrudge:
Can you share a simple project that calls this method twice? I can't repro here.

@chrisscott the `'auth0.is.authenticated'` key is misleading haha.. it should be `'auth0.has.the.user.authenticated.before.?'`

The idea is: if this browser didn't authenticate before (or if the cookie expired), then it makes no sense to call `getTokenSilently`, because it will most likely return `login_required`.

chrisscott:
Thanks for clarifying and suggesting a simple app to test. I did that and I'm unable to reproduce so it must be something happening in the app I've been testing with. Closing this out.

PR #13, opened 2019-04-29
Add default leeway of 60s

luisrudge:
This will prevent the 'token was issued in the future' error.


damieng:
Shouldn't the leeway calculation be:  Issued at - leeway/2  and Expires at +leeway/2 ?

Issue #24, opened 2019-05-23
handleRedirectCallback does not handle authorization error responses


nyobe:
It would be nice if `handleRedirectCallback` would extract error responses to the callback. I'm envisioning it throwing an error with with the `error` and `error_description` params for instance.

davidpatrick:
Thanks for reporting this @nyobe.  We will include this in our next release.

luisrudge:
released in @auth0/auth0-spa-js@1.0.0-beta.3 🎉

PR #25, opened 2019-05-29
Error handling in handleRedirectCallback

davidpatrick:
Throw errors provided by Auth0 authentication callback query params

Resolves https://github.com/auth0/auth0-spa-js/issues/24

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -397,7 +397,7 @@ describe('Auth0', () => {
           AuthenticationError
         );
       });
-      it('throws error with message from error_description', async () => {
+      it('throws AuthenticationError with message from error_description', async () => {
         const { auth0, utils } = await localSetup();
         const queryResult = {
           error: 'unauthorized',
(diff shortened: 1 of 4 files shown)


src/global.ts
```diff
@@ -177,12 +177,22 @@ interface LogoutOptions {
   client_id?: string;
 }
 
+/**
+ * @ignore
+ */
+interface AuthenticationErrorOptions {
```
luisrudge: what's this for?

src/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class AuthenticationError extends Error {
+  constructor(name: string, message: string) {
+    super(message);
+    this.name = name;
```
luisrudge: I think I rather have both `error` and `error_description` in the error.

```
super(error_description);
this.error = error;
this.error_description = error_description;
```

__tests__/index.test.ts
```diff
@@ -386,6 +388,27 @@ describe('Auth0', () => {
...
+      it('throws error with AuthenticationError', async () => {
+        const { auth0, utils } = await localSetup();
+        const queryResult = { error: 'unauthorized' };
+        utils.parseQueryResult.mockReturnValue(queryResult);
+
+        await expect(auth0.handleRedirectCallback()).rejects.toBeInstanceOf(
+          AuthenticationError
+        );
+      });
+      it('throws error with message from error_description', async () => {
```
luisrudge: there's no test to validate we're adding the `error` property to the AuthenticationError object

src/errors.ts
davidpatrick: Do you recommend not extending the Error class?  I've mapped `error_description` to `message` and `error` to the name.

luisrudge: We should absolutely extend the error class, but we should keep the original properties (`error` and `error_description`) intact so the dev can do:

```
catch ({error, error_description}) {
  //handle it
}
```

luisrudge: Also, if there's more information in the error that we receive (like `state` or something else), they should be here as well.

src/errors.ts
```diff
@@ -0,0 +1,12 @@
+export class AuthenticationError extends Error {
+  error: string;
+  error_description: string;
+  state: string;
+
+  constructor(error: string, error_description?: string, state?: string) {
```
luisrudge: 😬typescript tip of the day 😬 

you can add `pulic` or `private` in the constructor args to automatically create the fields in the class:

```js
class CustomError extends Error {
  constructor(
    public error: string,
    public error_description: string,
    public state: string
  ) {
    super(error_description);
  }
}

```

davidpatrick: Thanks @luisrudge !

PR #32, opened 2019-06-21
Reduce transaction cookie size

luisrudge:
### Description

Reduces the minimum transaction cookie size from `~455b` to `~288b`. If you use a custom `appState`, your cookie size will be bigger.

This improves the total cookie size from further requests. Most web servers have a `8192b` limit for cookies and will fail the request if you have more than that. This leaves us with ~28 available failed transactions. Each successful redirect callback will clear its own transaction, so only failed/cancelled transactions count for this limit.

### References

https://help.heroku.com/TQ80D553/why-do-i-get-a-400-bad-request-response-when-i-have-large-cookies

### Testing

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -328,15 +328,16 @@ describe('Auth0', () => {
       const { auth0, transactionManager } = await setup();
 
       await auth0.loginWithRedirect(REDIRECT_OPTIONS);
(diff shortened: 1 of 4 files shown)


luisrudge:
> Does it make sense to add a regression test to verify the size of the cookie and that it doesn’t exceed the limit?

I don't think so. If we need to add another piece of information in the future, we just need to make sure we absolutely need it.

Issue #97, opened 2019-07-19
Syntax Error on IE


izzmeifter:
The lib was erroring on our project in IE 11 after updating to the spa lib from lock, and when testing with the demo application we get the same error. It looks like it's using some es6 syntax in the module (spread operator). I'm using differential loading, but it doesn't look like it effects node_modules. below are the errors from the demo app.

This is on IE 11.0.9600 

Perhaps i'm doing something wrong in the configuration, but i can't find anything.

Thanks.
<img width="330" alt="Screen Shot 2019-07-19 at 3 47 11 PM" src="https://user-images.githubusercontent.com/4297079/61567707-c7017800-aa3d-11e9-92cb-80d9d593a605.png">
<img width="457" alt="Screen Shot 2019-07-19 at 3 56 08 PM" src="https://user-images.githubusercontent.com/4297079/61567709-c7017800-aa3d-11e9-8efc-f1f53ed0282d.png">

luisrudge:
Ops! That was an oversight of my part. I opened a PR to fix the file bundles will pick when you import our library. I'm so sorry about this. If you need a quick replacement for this weekend, feel free to use our CDN bundle that is ES5 compatible and has all the polyfills available: https://cdn.auth0.com/js/auth0-spa-js/1.1.0/auth0-spa-js.production.legacy.js - it's a bit bigger than the original one, but will work in IE11 just fine!

keithrz:
@luisrudge does auth0-spa-js.production.legacy.js add createAuth0Client to the global window object, just like auth0-spa-js.product.js does ?  I don't see how it does.  If not, how are you supposed to use auth0-spa-js.production.legacy.js ?  (this may be a moot point if a new build was generated with the latest code from master)

luisrudge:
@keithrz you're totally right. It didn't expose the object in the window 🤦‍♂ In any case, `legacy` is going away and we're shipping the IE11 polyfills in the main bundle in the next version that will be released next week.

PR #98, opened 2019-07-19
Make sure our production bundle is ES5 compatible.

luisrudge:
Fix https://github.com/auth0/auth0-spa-js/issues/97

### Description

Our production bundle was shipping ES2015 code, which is not supported in IE11. This PR fixes it and also adds a check to make sure our production build is ES5 compatible.


rollup.config.js
```diff
@@ -8,13 +8,12 @@ import livereload from 'rollup-plugin-livereload';
...
 
 import pkg from './package.json';
-import { Object } from 'core-js';
 
 const EXPORT_NAME = 'createAuth0Client';
 
 const isProduction = process.env.NODE_ENV === 'production';
 const shouldGenerateStats = process.env.WITH_STATS === 'true';
-const getPlugins = (shouldMinify, tsconfigOverride = {}) => {
+const getPlugins = (shouldMinify, { target }) => {
```
damieng: Is this a breaking change for people who were providing additional options?

echee:
will this also fix the IE11 issue with the way crypto is used in the production bundle?

luisrudge:
@echee using msCrypto? If you load it via CDN, you can use the legacy bundle which has this included. If you have a bundler, you're probably patching this yourself, right?

echee:
@luisrudge currently no, we are installing the npm package. We'd tried to polyfill this ourselves but thought surely it should be built in if the browser support is for IE11 too. Correct me if i'm wrong/

luisrudge:
@echee My first thought was to let devs consuming the package via npm to do it themselves, but I think you're right! We should provide this OOTB. I'll patch this in the next release. Thanks for bringing this up!

Issue #70, opened 2019-07-14
Getting an uncaught error in the react quick start example


mcyleung:
I am getting the following error in the download sample that you guys have provided for the react quick start of auth0
Uncaught (in promise) TypeError: Cannot read property 'split' of undefined
    at auth0-spa-js.production.js:685

luisrudge:
do you have a code snippet? what lib version is this?

mcyleung:
I think this bug has something to do with authorization from what I got from the logs and I am using "@auth0/auth0-spa-js": "^1.0.2". 
I don't really have a code snippet for you because this is straight from the quick start example. But I can DM you my domain and ClientID for you to try on the react quick-start sample.

What I have tried:
- It works fine when I have a fresh SPA application.
- It breaks on my old SPA application. I don't have any rules.

luisrudge:
do you have a HAR file I can take a look? You can send it to luis.rudge at auth0.com

luisrudge:
@mcyleung just sent me an e-mail with a log and with the following text:

> Unfortunately, I cannot reproduce it anymore and therefore I don't have a HAR file for you to look at.
> The authentication was a success, but the authorization failed.

If anyone hits this again, please open a new issue so we can investigate. Thanks @mcyleung for following up by email!

mkmuduli:
![image](https://user-images.githubusercontent.com/16792091/61880305-47c6e500-af12-11e9-8330-dfa491d9b4ef.png)

Getting the same issue. Trying to login using gmail. Login is succesful but on Approving access to the application, the page redirects to http://localhost:3000/?code=YpqazoJ6B_RazjFx&state=Mm05b19LSlZSekxaTGpnZGR1aHhBNnF%2BSkhiczZneHI3dm91anI3NEF5RA%3D%3D and gives the error in console. Check attached screeenshot

Kindly help resolve.

luisrudge:
Can you inspect why your oauth/token http request is failing? This will help you solve this issue. Maybe it's a rule or something?

zuercherm:
I'm having this issue as well. I cannot figure out why the token request is getting a 401. In the Auth0 logs it shows a "Failed Exchange" due to "Unauthorized". Is there anything specific I should look into to understand what's causing it? I have no rules FWIW.

Thanks

luisrudge:
@zuercherm Can you check if your client was created as an SPA client? I saw this happening with clients that weren't configured as SPA.

[...]

PR #116, opened 2019-07-26
Handle oauth/token request errors

luisrudge:
### Description

Our http request library doesn't throw an error if the request is 401, 403 etc, so it's up to us to implement the error handling logic. This PR adds logic to catch unsuccessful requests (non 2xx status code) and throws the appropriate error to inform developers what's going on.

### References

This greatly improves DX when dealing with issues like https://github.com/auth0/auth0-spa-js/issues/70

### Testing

### DIFF
--- a/__tests__/utils.test.ts
+++ b/__tests__/utils.test.ts
@@ -167,7 +167,7 @@ describe('utils', () => {
       (<any>global).fetch = jest.fn(
         () =>
           new Promise(res =>
-            res({ json: () => new Promise(ress => ress(true)) })
+            res({ ok: true, json: () => new Promise(ress => ress(true)) })
           )
       );
       await oauthToken({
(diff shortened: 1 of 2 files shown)


damieng:
Is this a breaking change?

Presumably at a min bar we need to update the docs and quickstarts to show error handling.

luisrudge:
@damieng no. before, it would just throw a cryptic error like this:
image

Now it will throw an actual method with error and error_description. This actually makes it work with any error handling you already had in place.

JohnnyTMD:
Hi, I still get this error.
I'm using the latest version, the token is ok. The error is thrown by this method auth0.handleRedirectCallback() 
I guess it is just a typo, the actual field name in the request is access_token, in the library I can see it's searching for id_token.
jwt.ts:37
Screenshot from 2019-08-29 21-46-22

luisrudge:
That's super weird. Means that the HTTP request is ok, but you're not getting the id_token 🤔 can you please send a HAR request to luis.rudge@auth0.com? I'll investigate.

JohnnyTMD:
I have solved the issue, @luisrudge helped me with that.
In case you wonder, the issue was caused by a custom rule, so I have disabled it and after that it worked.

luisrudge:
Perfect @JohnnyTMD. Can you figure it out what in the rule was removing the id_token from the response?

PR #129, opened 2019-08-02
Add ?federated logout param

### LINKED ISSUE #126: Federated logout
It seems like auth-spa-js 1.1.0 currently does not support telling auth0 to perform a [federated logout](https://auth0.com/docs/logout/guides/logout-idps). There seems to be no way to pass the required _federated_ query parameter since the `LogoutOptions` interface does not allow this:
https://github.com/auth0/auth0-spa-js/blob/5f946338af92183298d93161e9f181b823cd2ec2/src/global.ts#L162-L179

Am i missing something? If not, can you please add an option to support auth0's federated logout?

luisrudge:
fix #126 

### Description

Adds option to send the `federated` query param to /logout

### References

https://auth0.com/docs/logout/guides/logout-idps

### Testing

### DIFF
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -403,7 +403,9 @@ export default class Auth0Client {
       delete options.client_id;
     }
     ClientStorage.remove('auth0.is.authenticated');
-    const url = this._url(`/v2/logout?${createQueryParams(options)}`);
-    window.location.assign(url);
+    const { federated, ...logoutOptions } = options;
+    const federatedQuery = federated ? `&federated` : '';
+    const url = this._url(`/v2/logout?${createQueryParams(logoutOptions)}`);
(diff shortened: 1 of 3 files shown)


src/Auth0Client.ts
```diff
@@ -403,7 +403,9 @@ export default class Auth0Client {
       delete options.client_id;
     }
     ClientStorage.remove('auth0.is.authenticated');
-    const url = this._url(`/v2/logout?${createQueryParams(options)}`);
-    window.location.assign(url);
+    const { federated, ...logoutOptions } = options;
+    const federatedQuery = federated ? `&federated` : '';
+    const url = this._url(`/v2/logout?${createQueryParams(logoutOptions)}`);
```
damieng: Why wouldn't createQueryParams do this?

Issue #99, opened 2019-07-22
auth0_spa_js_1.default is not a function


Dig-Doug:
Hi, I'm trying to use the auth0 library in my project but I'm getting an error:

```
ts_scripts.js?v=1563805379409:34261 ERROR Error: Uncaught (in promise): TypeError: auth0_spa_js_1.default is not a function
TypeError: auth0_spa_js_1.default is not a function
```

I've seen some similiar issues, #60, #45, but their setups are a bit different and I've tried the suggested tsconfig fixes and I couldn't get them to work.

My project uses a setup similar to the [angular-bazel-example](https://github.com/angular/angular-bazel-example), so I've created a repro of the problem using it. Steps are below.

```bash
cd /tmp
git clone https://github.com/Dig-Doug/angular-bazel-example.git
cd angular-bazel-example
git checkout 8fa48dc1b977c435178aa5bfc1f9d601ed15f5b7
yarn install
npx bazel run //src:devserver

# It should say "Server listening on XXX"
```
The auth service from the example is in `src/app/auth/` and I've tried to use the service in `src/app.component.ts`.

When you connect to the local server, the page will load and lots of messages will be printed to the console. Make sure you scroll to the top, the error message is there.

I think the problem has something to do with how the `auth0` library is being exported or imported. Looking at the generated code, `auth.service.js` (note JS), it tries to access the library with:

```js
this.auth0Client = yield auth0_spa_js_1.default(this.config);
```

In my main project I use lots of other 3rd party libraries, e.g. `immutable`, `three`, without issue so it's not an issue with loading libraries in general.

Please let me know if you need any more info.

Thanks!

necramirez:
We have the same problem. We can't change our tsconfig as the other issues say because doing that breaks our build.

luisrudge:
ok, found the issue, although I can't really say why TS works like this 😬 

LOOOOTS of other libraries have the same issue: https://www.google.com/search?q=typescript+default+is+not+a+function

https://github.com/iamkun/dayjs/issues/475#issuecomment-483792717

> Ahh actually it was a bit more complicated I'm realizing then what I wrote. Basically when you have a library compiling with tsconfig allowSyntheticDefaultImports and then a consuming application that does not use that setting, we got that error. Both the library and consuming application had dayjs as dependencies.

[...]

PR #130, opened 2019-08-02
Add `default is not a function` entry to our FAQ

luisrudge:
fix https://github.com/auth0/auth0-spa-js/issues/99

### Description

Add new entry to our FAQ ❓ 

### References

https://github.com/auth0/auth0-spa-js/issues/99


FAQ.md
```diff
@@ -62,3 +62,7 @@ With this change, an immediate refresh after login works as expected.
 Note that even though the workaround doesn't cause any weird side effects in browers, you should ideally remove it after the bug has been fixed in Firefox.
 
 For more context see this [issue](https://github.com/auth0-samples/auth0-react-samples/issues/145).
+
+## Why do I get `auth0_spa_js_1.default is not a function` when using Typescript?
+
+Due to how the type system works in Typescript, if one of your dependencies uses `allowSyntheticDefaultImports: true`, then all the consumers of that dependency must use `allowSyntheticDefaultImports: true` as well. This, of course, is not ideal and might break your app if you depend on this setting being `false`. The Typescript team came up with a new flag that helps in this scenario. If you're hitting this issue, set `esModuleInterop: true` in your `tsconfig.json` (inside `compilerOptions`) file.
```
damieng: How about we put the solution first and then more info about it afterward.  I.e. start with

>If you're hitting this issue, set `esModuleInterop: true` in your `tsconfig.json` (inside `compilerOptions`) file.

Then "The reason for this is because how the type system works in TypeScript.."...

luisrudge: done

Issue #92, opened 2019-07-17
Configure or Handle Timeout on Popup Authentication


mwillbanks:
When using loginWithPopup: 
```
    let token;
    try {
      token = await auth0Client.getTokenWithPopup(params);
    } catch (error) {
      console.error(error);
    }
```

An issue arrises that when an error occurs such as leaving the window open until it times out there is not a recourse to shut the popup window, possibly this should be a parameter with the `getTokenWithPopup` where a parameter such as `closePopupOnError` or otherwise would exist.  We should also likely be able to control the timeouts associated with this rather than the hard coding to 30s inside of the util script.

This is a fairly large blocker for us but I am not certain how to handle this other than forking the current code and re-integrating it that way.

luisrudge:
I agree that will be super useful. Do you want to send a PR with that? If you can't, I'll add our near term backlog to make sure we work on this.

Enngage:
Same here, I feel the 30s timeout is way too low :( I would really liked there to be a way to configure custom timeout + way to close popup on `any` error.

luisrudge:
@mwillbanks @Enngage I'd love your feedback on #133 🎉

Enngage:
@luisrudge awesome, thank you! 

Could you possibly share the reasoning behind the default 30sec timeout? Many users fail to fill in their credentials in such short time. 

I like that you can close the popup from within the error using `error.popup.close();`, but would it be possible to get an instance of popup when you call `auth0.loginWithPopup` (and similar methods) so that we can potentially have more control and close popup whenever we want?

luisrudge:
@Enngage I'm not sure where you saw a 30s default. I think we always used a 60s timeout. 
It's not possible to return the popup in the `loginWithPopup` method because when the method finishes, the authentication already happened. I think it's better if we manage the lifecycle of the popup.

mwillbanks:
You can fetch the pop-up today it's a hack but works with most browsers: `const win = window.open('', 'auth0:authorize:popup')`

Enngage:
@luisrudge I thought I saw 30 seconds somewhere, but maybe I confused it with something else :-) Either way the PR looks good and would definitely help accomplish both my issues I have currently with   this package. Thank you!

luisrudge:
> You can fetch the pop-up today it's a hack but works with most browsers: const win = window.open('', 'auth0:authorize:popup')

[...]

PR #133, opened 2019-08-02
Add popup timeout config

luisrudge:
fix https://github.com/auth0/auth0-spa-js/issues/92

### Description

Adds a config object to send popup configuration options. For now, only `timeoutInSeconds` is available.

### References

https://github.com/auth0/auth0-spa-js/issues/92

### Testing


mwillbanks:
Looks great to me.

__tests__/utils.test.ts
```diff
@@ -297,13 +297,33 @@ describe('utils', () => {
...
       expect(popup.close).toHaveBeenCalled();
     });
-    it('times out after 60s', async () => {
+    it('times out after config.timeoutInSeconds', async () => {
+      const { popup, url } = setup('');
+      const seconds = 10;
+      /**
+       * We need to run the timers after we start `runPopup`, but we also
+       * need to use `jest.useFakeTimers` to trigger the timeout.
+       * That's why we're using a real `setTimeout`, then using fake timers
```
lbalmaceda: image

src/utils.ts
```diff
@@ -54,12 +54,16 @@ export const openPopup = () => {
...
+  authorizeUrl: string,
+  config: PopupConfigOptions
+) => {
   popup.location.href = authorizeUrl;
   return new Promise<AuthenticationResult>((resolve, reject) => {
     const timeoutId = setTimeout(() => {
-      reject(TIMEOUT_ERROR);
-    }, 60 * 1000);
+      reject({ ...TIMEOUT_ERROR, popup });
+    }, config.timeoutInSeconds || 60 * 1000);
```
lbalmaceda: I see you are using `60` here and in `src/Auth0Client.ts`. Please, find a way to reuse this definition, either via a constant or file.

src/global.ts
```diff
@@ -117,6 +117,9 @@ interface RedirectLoginResult {
 }
 
 interface PopupLoginOptions extends BaseLoginOptions {}
+interface PopupConfigOptions {
```
lbalmaceda: docs please, anywhere? Maybe on the README?

src/utils.ts
luisrudge: done

src/global.ts
luisrudge: done

__tests__/index.test.ts
```diff
@@ -32,6 +32,10 @@ const TEST_TELEMETRY_QUERY_STRING = `&auth0Client=${encodeURIComponent(
   )
 )}`;
 
+const DEFAULT_POPUP_CONFIG_OPTIONS: PopupConfigOptions = {
+  timeoutInSeconds: 60
```
lbalmaceda: Shouldn't this be pointing to the constant file as well?
```
import { DEFAULT_AUTHORIZE_TIMEOUT_IN_SECONDS } from '../src/constants';

```

src/global.ts
```diff
@@ -118,6 +118,14 @@ interface RedirectLoginResult {
 
 interface PopupLoginOptions extends BaseLoginOptions {}
 
+interface PopupConfigOptions {
+  /**
+   * The number of seconds to wait for a popup response before
+   * throwing a timeout error
```
lbalmaceda: Please add the default value, or point to the constant name for the user to check if you think this will change often.

PR #154, opened 2019-08-16
Add IE11 polyfills

### LINKED ISSUE #113: Polyfills needed for IE 11?
Hey 👋 

I have a Vue CLI app that I've adding this package to via npm. I have it all working in modern browsers, but IE 11 throws a number of errors related to missing features (i.e., `String.prototype.startsWith`).

I could be totally off here, but it looks like `dist/auth0-spa-js.production.js` is built to be ES5 compatible, and work in IE 11?

If not, my follow-up question: What are the polyfills this package needs for IE 11?

Vue CLI [recommends](https://cli.vuejs.org/guide/browser-compatibility.html#polyfills) either explicitly including the polyfills needed (which I've done with other dependencies), or just including all polyfills for your target browsers. I'd rather not include all polyfills if possible.

luisrudge:
### Description

- Add IE11 polyfills
- remove `legacy.ts` file and `.legacy.js` bundle file. All bundles will be IE11 compatible, so there's no need to have a separate bundle with polyfills
- replace `qs` by `qss` (saves 3kb gzipped)
- make index.html file IE11 compatible
- make all bundles ES5 compatible
- stop using `Set` so we don't have to polyfill it (it's a big polyfill)

### References

Fix #113 

### Testing


src/index.ts
```diff
@@ -1,18 +1,32 @@
...
 //this is necessary to export the type definitions used in this file
 import './global';
 
 export default async function createAuth0Client(options: Auth0ClientOptions) {
+  if (!window.crypto && (<any>window).msCrypto) {
+    (<any>window).crypto = (<any>window).msCrypto;
+  }
+  if (!window.crypto) {
+    throw new Error(
+      'For security reasons, `window.crypto` is required to run `auth0-spa-js`.'
```
lbalmaceda: Isn't this breaking?

luisrudge: No. This was already inside our `legacy` bundle. I just moved it here because we're now shipping everything IE11 compatible. <IE10 doesn't have window.crypto available, so it wouldn't work anyway.

keithrz:
Great, a fix to IE11 polyfills!  Now that it is in master, can we get a new version of auth0 built & packaged with this fix?

luisrudge:
@keithrz next week 🎉

Issue #157, opened 2019-08-18
Export Auth0Client typing


jcmoore0:
Hi Auth Team - I would like to be able to import the Auth0Client like this (standard practice)
```
import createAuth0Client, { Auth0Client } from '@auth0/auth0-spa-js';
```
Would you please add to src/index.d.ts?
```
export { Auth0Client} 
```
Auth0Client is imported but not exported in src/index.d.ts.
Thanks for the help.

luisrudge:
closing for this reason: https://github.com/auth0/auth0-spa-js/pull/159#issuecomment-522601580

PR #159, opened 2019-08-19
Expose Auth0Client, Auth0ClientOptions from index.ts

paul-thompson-helix:

### Description

* `export { Auth0Client, Auth0ClientOptions };` from `index.ts` for improved typescript typings.
* Required fixing exports on the `global.ts` file and imports of that.
* Required fixing the rollup to bundle from a new bundle.ts file with the `export default createAuth0Client;` previously in index.ts
* Check tests pass

### References

https://github.com/auth0/auth0-spa-js/issues/157

### Testing
`npm run test`
`npm run test:integration`

### Checklist

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -4,6 +4,8 @@ jest.mock('../src/cache');
 jest.mock('../src/transaction-manager');
 jest.mock('../src/utils');
 
+import { PopupConfigOptions, GetTokenSilentlyOptions } from '../src/global';
+
 import Auth0Client from '../src/Auth0Client';
 import createAuth0Client from '../src/index';
 import { AuthenticationError } from '../src/errors';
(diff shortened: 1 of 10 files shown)


luisrudge:
Thanks for the PR but it's unlikely we'll be exposing the internals of the SDK. This would make it harder to implement changes because every change in the internals would be considered a breaking change. At the end of the day, the goal of this SDK is to be opinionated in how you should use it, so we'll stick with the current public API for now.

PachowStudios:
I seriously disagree with the choice to not do this. createAuth0Client returns Auth0Client, effectively making it public API, and therefore should be exported. Not sure how you consider Auth0Client to be "internals" when your own docs literally tell us to import it via the unwieldy `import Auth0Client from '@auth0/auth0-spa-js/dist/typings/Auth0Client';`.

This goes against how every other TypeScript library out there does things.

If you really don't want to return the class, then please export an interface that the class implements.

seeruk:
Any chance the above comment could be reviewed? This code is pretty gross:

```ts
import Auth0Client from "@auth0/auth0-spa-js/dist/typings/Auth0Client"
```

When it could be imported alongside `createAuth0Client`:

```ts
import createAuth0Client, { Auth0Client } from "@auth0/auth0-spa-js"
```

Literally the only thing that would change is that it'd make this library nicer to use, nothing else.

PR #197, opened 2019-09-02
Fix incorrect state extraction from query string

### LINKED ISSUE #186: Always "Invalid state" in v1.2
Hey 👋 

I just tried upgrading from v1.1.1 to v1.2 (doesn't matter which patch version), and I now get an "Invalid state" error 100% of the time when calling `handleRedirectCallback`. I've tried in both Chrome and Firefox, cleared cookies, etc. I didn't change any code – just the version number in `package.json`.

I'm happy to provide more details, I'm just not sure what would be helpful to know in this case. Let me know. Thanks!

luisrudge:
### Description

CLIs that are aggressively trying to minify our already minified code (like vue-cli and ng-cli) are causing issues with one of our dependencies (`qss`).

What happens is that the second minification pass modifies the content or our dependency making it run a different function.
image

Because of this bug, when our code tries to get the query string from the URL, it doesn't find the `state` param and fails.

This PR removes `qss` as dependency and just encodes/decodes query strings manually.

`qss` is still used as a dev dependency for integration tests.

### References

Fix #186 

### Testing

No test change because we were already testing the result of the parsed/encoded results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "pem": "^1.14.2",
     "prettier": "^1.18.2",
     "pretty-quick": "^1.11.1",
+    "qss": "^2.0.3",
     "rimraf": "^3.0.0",
     "rollup": "^1.17.0",
     "rollup-plugin-commonjs": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "@auth0/component-cdn-uploader": "auth0/component-cdn-uploader#v2.2.2",
     "@types/cypress": "^1.1.3",
     "@types/jest": "^24.0.15",
-    "@types/qs": "^6.5.3",
     "@typescript-eslint/eslint-plugin-tslint": "^2.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -50,6 +49,7 @@
     "pem": "^1.14.2",
     "prettier": "^1.18.2",
     "pretty-quick": "^1.11.1",
+    "qss": "^2.0.3",
```
damieng: Do we still need qss then?

luisrudge: `qss` is still used as a dev dependency for integration tests.

src/utils.ts
```diff
@@ -14,7 +13,13 @@ export const getUniqueScopes = (...scopes: string[]) => {
 };
 
 export const parseQueryResult = (hash: string) => {
```
damieng: I'd rename this fragment or something to avoid name conflict with the hash parameter on line 18.

luisrudge: pushed

PR #230, opened 2019-09-30
Add missing char for nonce/state generation.

gkwang:
### Description
We're missing a **W** in the nonce/state generator. 
This does not impact security in any significant way; making this change so it has the expected behavior.


gkwang:
oops, let me fix the tests.

PR #232, opened 2019-10-02
Fix safari10 initialization error

luisrudge:
### Description

Safari 10 uses `crypto.webkitSubtle` instead of `crypto.subtle`. This PR normalizes this behavior.

### References

ESD-2220

### Testing

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,28 +10,10 @@ import * as ClientStorage from './storage';
 
 //this is necessary to export the type definitions used in this file
 import './global';
+import { validateCrypto } from './utils';
 
 export default async function createAuth0Client(options: Auth0ClientOptions) {
-  // ie11 compat
-  if (!window.crypto && (<any>window).msCrypto) {
-    (<any>window).crypto = (<any>window).msCrypto;
-  }
-  if (!window.crypto) {
-    throw new Error(
-      'For security reasons, `window.crypto` is required to run `auth0-spa-js`.'
-    );
-  }
-  // safari 10 compat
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -20,10 +21,14 @@ export default async function createAuth0Client(options: Auth0ClientOptions) {
       'For security reasons, `window.crypto` is required to run `auth0-spa-js`.'
     );
   }
+  // safari 10 compat
+  if ((<any>window.crypto).webkitSubtle) {
```
damieng: The guidance at https://webkit.org/blog/7790/update-on-web-cryptography/ suggests you use subtle first and only fall back to webkitSubtle if that fails.  It looks like this is reversed?  If it exists it uses it over subtle?  They specifically call out that as bad behavior on that page.

luisrudge: fixed

luisrudge:
@damieng updated 🎉 thanks for the reference

src/index.ts
```diff
@@ -20,10 +21,14 @@ export default async function createAuth0Client(options: Auth0ClientOptions) {
       'For security reasons, `window.crypto` is required to run `auth0-spa-js`.'
     );
   }
+  // safari 10 compat
+  if (!window.crypto.subtle && (<any>window.crypto).webkitSubtle) {
+    (<any>window.crypto).subtle = (<any>window.crypto).webkitSubtle;
```
damieng: No non-Auth0 code can be running in the Window while we're doing this right?

Trying to decide if replacing a global is safe in this context or whether we should be creating our own variable to use instead to avoid non-Auth0 code getting confused.

PR #233, opened 2019-10-03
Add early expiration of Access Token in cache

B-Galati:
### Description

Hello, 

**Problem:** In my case the access token provided by Auth0 is valid for 30 seconds and from time to time I would have 401 when trying to call the resource server.

image

**Proprosed solution:** this PR expires the access token 5 seconds before it really expires so that we can be more sure 
the access token used against the resource server is not expired by the time it reaches it.

Let me know what you think.

### Testing

### Checklist

### DIFF
--- a/__tests__/cache.test.ts
+++ b/__tests__/cache.test.ts
@@ -54,7 +54,9 @@ describe('cache', () => {
         user: { name: 'Test' }
       }
     });
-    jest.advanceTimersByTime(1001);
+    jest.advanceTimersByTime(799);
+    expect(Object.keys(cache.cache).length).toBe(1);
+    jest.advanceTimersByTime(1);
     expect(Object.keys(cache.cache).length).toBe(0);
   });
   it('expires after `user.exp` when `user.exp` < `expires_in`', () => {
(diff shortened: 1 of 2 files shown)


luisrudge:
@B-Galati thanks for the PR. Can you clarify how are you using the `getTokenSilently` method? Are you calling it *right before* doing the API call? Or are you storing the token somewhere else etc? 

Regarding your PR, 5s is not the best solution. If we decide to move forward with this, I think we need to use a percentage based approach. Something like `Math.min(expiresIn, expTime) * 1000 * 0.8;`, so when it gets to 80% of the expiration time. Can yo please do this change and write a test? Thanks!

B-Galati:
@luisrudge Good idea! PR updated along with tests.

>Can you clarify how are you using the getTokenSilently method? Are you calling it right before doing the API call? Or are you storing the token somewhere else etc?

That's exactly it, calling it right before doing the API call using the cache strategy provided by the library.

luisrudge:
Thank you so much for this PR!

B-Galati:
Thank you for the quick merge!

Issue #109, opened 2019-07-24
Unable to call getTokenSilently concurrently


Tazaf:
Hi!

I'm having an error in the console when creating the Auth0 client using `createAuth0Client()`.

The error is the following:

```
> ERROR TypeError: Cannot read property 'close' of undefined
    at o (auth0-spa-js.production.js:formatted:1)
    at ZoneDelegate.invokeTask (zone-evergreen.js:391)
    at Object.onInvokeTask (auth0-spa-js.production.js:formatted:1)
    at ZoneDelegate.invokeTask (zone-evergreen.js:390)
    at Zone.runTask (zone-evergreen.js:168)
    at ZoneTask.invokeTask [as invoke] (zone-evergreen.js:465)
    at invokeTask (zone-evergreen.js:1603)
    at globalZoneAwareCallback (zone-evergreen.js:1640)
```
On the beautyfied production code, the error points to this line:
```js
o = function(i) {
  // Below is the culprit, apparently, 'i.source' is undefined
  i.origin == m && i.data && "authorization_response" === i.data.type && (i.source.close(),
  i.data.response.error ? t(i.data.response) : e(i.data.response),
  clearTimeout(n),
  window.removeEventListener("message", o, !1),
  window.document.body.removeChild(r))
};
```

And the code that's calling the `createAuth0Client()` function is this one (pretty much the exact same as the [Angular Login Tutorial](https://auth0.com/docs/quickstart/spa/angular2#add-an-authentication-service)):

```typescript
// This is a method of my AuthService
async getAuth0Client(): Promise<Auth0Client> {
  if (!this.auth0Client) {
    console.log("AuthService - generating new Auth0Client");
    this.auth0Client = await createAuth0Client(environment.auth0);
    console.log("AuthService - new Auth0Client", this.auth0Client);

    // Provide the current value of isAuthenticated
    this.isAuthenticated.next(await this.auth0Client.isAuthenticated());

    // Whenever isAuthenticated changes, provide the current value of `getUser`
    this.isAuthenticated
      .subscribe(async isAuthenticated => await this.updateUserProfile(isAuthenticated));
  }
  return this.auth0Client;
}
```

Note that this error does not prevent the client to be effectively created and functionning. But it's still an unwanted error 😄 

[...]

PR #238, opened 2019-10-10
Add 'lock' to prevent `getTokenSilently` to be invoked in parallel

luisrudge:
### Description
Today, the `getTokenSilently` method behaves like this:
- create an iframe
- attach a listener to the window to receive messages from the iframe
- when the message from the iframe arrives, it removes the listener and the iframe from the DOM

The problem we have today is that if you call the method concurrently, we can hit a race condition where, by having two iframes at the same time, the order of the response is mixed and by the time you try to remove the iframe, it’s already gone and it throws an error (https://github.com/auth0/auth0-spa-js/issues/109).

This PR implements a 'lock' the current instance of the Auth0Client in all instances of your app (from the same browser) until the `getTokenSilently` call is finished. This enables us to always* execute the method once at a time.

* This is not a real lock. It's a *best effort* lock due to platform limitations. In mobile or when a tab is 'sleeping', it might be the case that the lock is not made correctly for that particular tab, that's when proper error handling (something we'll work on) will help.

### References
fix https://github.com/auth0/auth0-spa-js/issues/109

### Testing


src/Auth0Client.ts
```diff
@@ -1,3 +1,5 @@
+import Lock from 'browser-tabs-lock';
```
joshcanhelp: Don't you mean:

```js
import Lock from 'auth0-lock';
```

😆

luisrudge: geez. can you imagine that? 😝

PR #242, opened 2019-10-14
Fix IE msCrypto.subtle usage

### LINKED ISSUE #241: createAuth0Client promise rejects with an error on Internet Explorer
The error is:

```
unable to get property ‘subtle’ of undefined...
```

https://github.com/auth0/auth0-spa-js/blob/78075fe8d38ab735b954209afb2532d8e90749fc/src/utils.ts#L187

Thanks for keeping IE alive! 😄

luisrudge:
### Description

When abstracting the `crypto` and `subtle` calls, we introduced a bug that prevented IE11 to get the correct instance of `subtle` (which comes from `msCrypto` and not `crypto`. This PR fixes this by using our internal abstraction `getCrypto` which already takes that into consideration.

### References

Fix #241 

### Testing

### DIFF
--- a/__tests__/utils.test.ts
+++ b/__tests__/utils.test.ts
@@ -527,6 +527,13 @@ describe('utils', () => {
       const theSubtle = getCryptoSubtle();
       expect(theSubtle).toBe('window');
     });
+    it('should use msCrypto.subtle when available', () => {
+      (<any>global).crypto = undefined;
+      (<any>global).msCrypto = { subtle: 'ms' };
+
+      const theSubtle = getCryptoSubtle();
+      expect(theSubtle).toBe('ms');
+    });
   });
   describe('validateCrypto', () => {
     it('should throw error if crypto is unavailable', () => {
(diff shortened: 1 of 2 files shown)


__tests__/utils.test.ts
```diff
@@ -527,6 +527,13 @@ describe('utils', () => {
       const theSubtle = getCryptoSubtle();
       expect(theSubtle).toBe('window');
     });
+    it('should use msCrypto.subtle when available', () => {
+      (<any>global).crypto = undefined;
+      (<any>global).msCrypto = { subtle: 'ms' };
+
+      const theSubtle = getCryptoSubtle();
```
joshcanhelp:
```suggestion
      const cryptoSubtle = getCryptoSubtle();
```

__tests__/utils.test.ts
```diff
@@ -527,6 +527,13 @@ describe('utils', () => {
...
+      expect(theSubtle).toBe('ms');
```
joshcanhelp:
```suggestion
      expect(cryptoSubtle).toBe('ms');
```

PR #248, opened 2019-10-17
Improved OIDC compliance

luisrudge:
### Description

This update improves the SDK support for OpenID Connect. In particular, it modifies the sign in verification phase by substituting backchannel based checks with id_token validation.

### Testing

### DIFF
--- a/src/jwt.ts
+++ b/src/jwt.ts
@@ -34,7 +34,12 @@ const idTokendecoded = [
 ];
 
 export const decode = (token: string) => {
-  const [header, payload, signature] = token.split('.');
+  const parts = token.split('.');
+  const [header, payload, signature] = parts;
+
+  if (parts.length !== 3 || !header || !payload || !signature) {
+    throw new Error('ID token could not be decoded');
+  }
   const payloadJSON = JSON.parse(urlDecodeB64(payload));
   const claims: IdToken = { __raw: token };
   const user: any = {};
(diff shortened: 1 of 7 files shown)


src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
...
+        );
+      }
+    }
+  } else if (decoded.claims.aud !== options.aud) {
+    throw new Error(
+      `Audience (aud) claim mismatch in the ID token; expected "${options.aud}" but found "${decoded.claims.aud}"`
+    );
+  }
+
+  if (!decoded.claims.nonce) {
```
lbalmaceda: should be optional unless your sdk ensures a value is always sent, even when user sets "null/undefined"

```suggestion
  if (options.nonce && !decoded.claims.nonce) {
```

src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
...
+    throw new Error(
+      'Nonce (nonce) claim must be a string present in the ID token'
+    );
+  }
+
   if (decoded.claims.nonce !== options.nonce) {
```
lbalmaceda: this one should be wrapped within `if (options.nonce)` (see sample implementation for reference)

src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
...
+  }
+
+  if (options.max_age && !decoded.claims.auth_time) {
+    throw new Error(
+      'Authentication Time (auth_time) claim must be a number present in the ID token when Max Age (max_age) is specified'
+    );
+  }
+
+  /* istanbul ignore next */
+  if (!decoded.claims.exp) {
```
lbalmaceda: check it's a number as well (applies to all date claims)

src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
...
+      'Issued At (iat) claim must be a number present in the ID token'
+    );
   }
 
   const now = new Date();
   const expDate = new Date(0);
   const iatDate = new Date(0);
   const nbfDate = new Date(0);
   const leeway = options.leeway || 60;
+  const authTimeDate = new Date(
```
lbalmaceda: why is this one being instantiated like this and the rest are using the `setUTCSeconds`? Seems all should be using the latter

src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
...
+      `Issued At (iat) claim error in the ID token; current time (${now}) is before issued at time (${iatDate})`
+    );
   }
   if (typeof decoded.claims.nbf !== 'undefined' && now < nbfDate) {
-    throw new Error('token is not yet valid (invalid notBefore)');
+    throw new Error(
+      `Not Before time (nbf) claim in the ID token indicates that this token can't be used just yet. Currrent time (${now}) is before ${nbfDate}`
+    );
+  }
+  if (typeof decoded.claims.auth_time !== 'undefined' && now > authTimeDate) {
```
lbalmaceda: Wouldn't it be more readable if you wrap this inside a presence-check if, like the one on line 140 ?

src/jwt.ts
```diff
@@ -53,37 +58,134 @@ export const decode = (token: string) => {
 };
 
 export const verify = (options: JWTVerifyOptions) => {
+  if (!options.id_token) {
```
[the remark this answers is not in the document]
luisrudge: this sdk always asks for id_token

src/jwt.ts
luisrudge: fixed

src/jwt.ts
luisrudge: but I only want to do the `now > authTimeDate` check if auth_time is present, so it's not the same check

src/jwt.ts
luisrudge: fixed

src/jwt.ts
```diff
@@ -53,37 +60,136 @@ export const decode = (token: string) => {
...
+      'Expiration Time (exp) claim must be a number present in the ID token'
+    );
+  }
+  if (!isNumber(decoded.claims.iat)) {
+    throw new Error(
+      'Issued At (iat) claim must be a number present in the ID token'
+    );
   }
 
+  const leeway = options.leeway || 60;
```
esarafianou: why do we set default leeway to 60 seconds? There is no mention about it in our docs.

gkwang: We decided to set the default leeway to 60 across the board to keep the consistency.

esarafianou: Makes sense. Let's include this piece of information here: https://auth0.github.io/auth0-spa-js/interfaces/auth0clientoptions.html#leeway

PR #376, opened 2020-03-16
[SDK-1127] Delay removal of iframe to prevent Chrome hanging status bug #240

adamjmcgrath:
### Description

Workaround for a Chrome browser bug where the loading status intermittently hangs when the `/authorize` iframe is removed in `getTokenSilently`

### References

- #240
- Also https://github.com/auth0/auth0.js/issues/1044

Screenshot 2020-03-16 at 11 19 49

### Testing

- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist

- ~[ ] I have added documentation for new/changed functionality in this PR or in auth0.com/docs~
- [x] All active GitHub checks for tests, formatting, and security are passing
- [x] The correct base branch is being used, if not `master`

PR #399, opened 2020-03-31
Check if iframe is still in body before removing

paulfalgout:
If the iframe is removed from the DOM prior to the timeout it would error on removeChild.

Error thrown: `Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

Bug introduced in https://github.com/auth0/auth0-spa-js/pull/376


adamjmcgrath:
Hi @paulfalgout,  thanks for raising this.

I'm not able to reproduce the error. Do you have a sample app I can take a look at? Also, what OS/browser are you using?

paulfalgout:
So it seems to happen on any OS/browser which I think makes sense.. Tried it on OSX Chrome and Firefox and Win10/IE11.  

I can have a live version once our next release goes out but it'd be hard for me to setup a repo to reproduce I think... although if there's a sample auth0 app I could modify easily I'm sure I could make it work..

Essentially though all that needs to happen is this:
```js
auth0.getSilentToken();
document.body.removeChild(document.getElementsByTagName('iframe')[0]);
// Then setTimeout occurs which attempts to re-remove the child
``` 

In my case, the iframe is appended to the body, and upon login success I clear out the `<body>` tag... so then the timeout to remove the iframe occurs after I've already cleared the body.

paulfalgout:
firefox
chrome
ie11

stevehobbsdev:
> although if there's a sample auth0 app I could modify easily I'm sure I could make it work

@paulfalgout Which stack are you using? We have sample apps for [React](https://github.com/auth0-samples/auth0-react-samples), [Vue](https://github.com/auth0-samples/auth0-vue-samples), [Angular](https://github.com/auth0-samples/auth0-angular-samples) and [plain JS](https://github.com/auth0-samples/auth0-javascript-samples) if they would help to produce something that demonstrates the error?

paulfalgout:
https://github.com/paulfalgout/auth0-javascript-samples/commit/09ea3d8bc154039dee1fbc62f2d81564e6e0cfeb#diff-343625e68d757fc5388fcd3955d10ddeR79

I would have had it much sooner, but I didn't realize until just now that I was trying to reproduce this problem on the hard-coded v1.2 in the sample repo 😢

paulfalgout:
@stevehobbsdev need anything else here?

adamjmcgrath:
Hi @paulfalgout - thanks for providing the example

I can see why removing the iframe immediately after logging in can throw the error, I'm just not sure why it's a valid use case. Could you just refrain from removing it?

> upon login success I clear out the `<body>` tag

Could you render your UI in a wrapper div, then clear that out on login success?

paulfalgout:
I don't see why clearing the `<body>` would be invalid generally.  It's typically recommended against doing so because 3rd party scripts like google fonts or browser extensions may want the body to inject into it ([ref](https://github.com/facebook/create-react-app/issues/1568)).  In my situation we're building a high security HIPAA app in which we generally _cannot_ use 3rd party scripts and in which we prefer to prevent browser extensions from injecting, so while trivial, it is a smallish security suggestion, to clear the body on load in these situations.

The _only_ reason this is a problem in this library is because a timeout was introduced.  Are you concerned that adding that single line `window.document.body.contains(iframe)` to check if it's still there is somehow a performance issue?  Or why would you _not_ want to add the check?

adamjmcgrath:
Hi, thanks for the explanation - we'll go ahead and merge this, bear with us while we organise the next release

PR #540, opened 2020-08-04
Update README.md

adamjmcgrath:
1.9 -> 1.11


lbalmaceda:
@adamjmcgrath I think the `release` script broke since we made the cdn link version specific to minors instead of patches. See https://github.com/auth0/auth0-spa-js/blob/master/scripts/release.js#L33-L36

PR #558, opened 2020-08-26
[SDK-1912] Unnecessary latency in `getTokenSilently` with primed cache

### LINKED ISSUE #552: getTokenSilently is slow, even with the cache
### Describe the problem

Calls to `getTokenSilently()` are slow, on the order of 50ms. In our case this breaks a subsequent use of [`apollo-link-batch-http`](https://www.apollographql.com/docs/link/links/batch-http/), as the delay pushes every query outside of the batching window.

The extra time comes from acquiring a lock to avoid multiple requests. The implementation was changed to fix this in https://github.com/auth0/auth0-spa-js/pull/339, but reverted in https://github.com/auth0/auth0-spa-js/pull/525 since it seems to have caused other unwanted behaviour.

Calls should only take a few milliseconds, as they do with `getIdTokenClaims`.

### Reproduction

1.  In the SDK Playground's `index.html`, replace the implementation of `showAuth0Info` with:
```js
        showAuth0Info: function () {
          var _self = this;
          _self.access_tokens = [];

          _self.auth0.getTokenSilently().then(function (token) {
            _self.audienceScopes[0].access_tokens.push(token);

            _self.auth0.getUser().then(function (user) {
              _self.profile = user;
            });

            _self.auth0.getIdTokenClaims().then(function (claims) {
              _self.id_token = claims.__raw;
            });
          }).then(() => {
            const start = performance.now();

            _self.auth0.getIdTokenClaims().then(() => {
              console.log(`getIdTokenClaims took ${performance.now() - start}ms`);
            });

            _self.auth0.getTokenSilently().then(() => {
              console.log(`getTokenSilently took ${performance.now() - start}ms`);
            });
          });
        },
```

2. Login via "Login popup" a few times.

3. Note the difference in processing times for the two methods. In our app we've seen a range of roughly 15ms - 90ms depending on whatever else the app is doing at the time.

```
getIdTokenClaims took 10ms
getTokenSilently took 81ms

getIdTokenClaims took 7ms
getTokenSilently took 79ms

getIdTokenClaims took 7ms
getTokenSilently took 67ms

getIdTokenClaims took 10ms
getTokenSilently took 80ms

getIdTokenClaims took 10ms
getTokenSilently took 77ms
```

(Currently on package version 1.11.0)

Cheers.

adamjmcgrath:
### Description

Check the cache before acquiring lock to avoid latency when the cache is populated (see #552)
Check the cache after acquiring lock to prevent unnecessary requests, because a previous call might have populated the cache while the current call was waiting to acquire the lock (see #480)

### References

fixes #552

### Testing

To check #552:

Call `getTokenSilently` with a primed cache, latency should be <20ms

Check regression of #480:

Hammer the `getTokenSilently` button in the SPA js playground with "Use token cache when fetching new tokens" enabled and check that only one token is requested 

### Checklist

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -1,12 +1,13 @@
-jest.mock('browser-tabs-lock');
+import { expectToHaveBeenCalledWithAuth0ClientParam } from './helpers';
+import { CacheLocation, Auth0ClientOptions } from '../src/global';
+import * as scope from '../src/scope';
+// @ts-ignore
+import { acquireLockSpy, releaseLockSpy } from 'browser-tabs-lock';
+
 jest.mock('../src/jwt');
 jest.mock('../src/storage');
 jest.mock('../src/transaction-manager');
 jest.mock('../src/utils');
-
-import { expectToHaveBeenCalledWithAuth0ClientParam } from './helpers';
-import { CacheLocation, Auth0ClientOptions } from '../src/global';
-import * as scope from '../src/scope';
 
 import createAuth0Client, {
   Auth0Client,
@@ -64,12 +65,6 @@ const webWorkerMatcher = expect.objectContaining({
 });
 
 const setup = async (clientOptions: Partial<Auth0ClientOptions> = {}) => {
-  const auth0 = await createAuth0Client({
-    domain: TEST_DOMAIN,
-    client_id: TEST_CLIENT_ID,
-    ...clientOptions
-  });
-
   const getDefaultInstance = m => require(m).default.mock.instances[0];
 
   const storage = {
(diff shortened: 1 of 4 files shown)


__tests__/index.test.ts
```diff
@@ -155,7 +154,7 @@ describe('Auth0', () => {
   });
 
   afterEach(() => {
-    jest.resetAllMocks();
+    jest.clearAllMocks();
```
adamjmcgrath, line 157: `resetAllMocks` destroys the `localStorage` mock (which browser-tabs-lock uses in these tests) https://github.com/clarkbw/jest-localstorage-mock/issues/83

Had to fix some leaky tests after removing it

src/Auth0Client.ts
```diff
@@ -582,22 +582,36 @@ export default class Auth0Client {
...
-          {
-            scope: getTokenOptions.scope,
-            audience: getTokenOptions.audience || 'default',
-            client_id: this.options.client_id
-          },
-          60 // get a new token if within 60 seconds of expiring
-        );
-
-        if (cache && cache.access_token) {
-          await lock.releaseLock(GET_TOKEN_SILENTLY_LOCK_KEY);
```
adamjmcgrath, line 599: This will be called in the `finally` so shouldn't be called here

stevehobbsdev (approved):

PR #779, opened 2021-08-19
Correct cache interface

employee451:
### Description

Corrects the `ICache` interface to include `null` as a possible return type of the `get` function.


stevehobbsdev (approved):
Thanks! 👍🏻

PR #793, opened 2021-09-12
add Auth0Client.getIdToken()

tgerk:
Auth0-issued id_tokens are useful in their original form & should be easy to get to.

### Description

Added a method to make it as straightforward to get the id_token field from the cache as it is to get the claims decoded from it.

### References

related to **issue 262** on @auth0/auth0-react

### Testing

- [√] This change adds test coverage for new/changed/fixed functionality

tests passed using node v14.15.0, jest@26.6.3
it was necessary to update tsconfig.json to accept es2015 and dom syntax/symbols

### Checklist

- [√ via docblock] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [√ npm build test] All active GitHub checks for tests, formatting, and security are passing
- [√] The correct base branch is being used, if not `master`


tgerk:
Not sure about "unfortunately"  I was sitting on the PR (here and in related auth0-react repo) since July.  Thanks to @khitrenovich comment, we got your feedback.

PR #1115, opened 2023-07-13
Usage nit

reify-tanner-stirrat:
This caught my eye in the readme.


Widcket (approved):
Thanks @reify-tanner-stirrat!

PR #1292, opened 2024-08-14
feat(getTokenWithPopup): Improve popup open error

ArthurKnoep:
### Changes

This PR introduces a new error class for the `getTokenWithPopup` method: `PopupOpenError`. This error will be thrown if the method is unable to open the popup (most likely because it has been blocked by the browser).

Having a specific error class makes it much easier to catch the error and do specific error handling (such as explaining to the users how to allow popups on their browser)

This PR should be retro-compatible with all developpers running a string comparison with the error message to detect popup permission issues as it uses the same error message.

### Testing

### Checklist

### DIFF
--- a/__tests__/Auth0Client/getTokenWithPopup.test.ts
+++ b/__tests__/Auth0Client/getTokenWithPopup.test.ts
@@ -8,6 +8,7 @@ import * as scope from '../../src/scope';
 import {
   assertPostFn,
   fetchResponse,
+  setupFailingPopup,
   setupFn,
   setupMessageEventLister
 } from './helpers';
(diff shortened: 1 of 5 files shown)


gyaneshgouraw-okta (approved):

gyaneshgouraw-okta:
Hi @ArthurKnoep, while I appreciate the goal of better error handling, this change is not backward compatible as claimed.
In my understanding this is how it may cause breaking error.
```
Before: throw new Error() → error.name = "Error"
After: throw new PopupOpenError() → error.name = "PopupOpenError"
```

While the error message remains the same, the error structure and type have changed, which will break existing error handling code that relies on these properties.

Could you please modify the approach to maintain true backward compatibility while still achieving the improvements?

yogeshchoudhary147 (approved):
This is a great DX improvement! Moving to a specific PopupOpenError is much cleaner and allows for robust instanceof checks in catch blocks.

Well done for keeping the error message identical to maintain full backward compatibility. LGTM!

PR #1348, opened 2025-03-14
fix: wrap `domainUrl` with `new URL().origin`

tonyfromundefined:
### Changes

- We're currently working on an OIDC provider that is compatible with Auth0 SDK, which will allow us to refine the experience for developers to easily integrate with the Auth0 SDK.
- Where our OIDC provider differs slightly from Auth0 is that it includes a subpath in the domain.
  - There is an issue where the `getAccessTokenSilently()` API does not work properly in this case.
- This shouldn't be an issue if it restricted to the Auth0 environment, but it would be semantically correct to wrap the `domainUrl` in `new URL(domainUrl).origin` rather than passing it directly in the code, so we're leaving a fix request.

### Testing

### Checklist


tusharpandey13 (approved):
lgtm

PR #1385, opened 2025-07-30
bugfix: Correctly handle refresh token rotation

tusharpandey13:
This change ensures that when a new refresh token is received due to rotation, all cache entries with the old refresh token are invalidated. This prevents the SDK from using a stale, rotated refresh token on subsequent token requests.

- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [x] I have added documentation for all new/changed functionality (or N/A)

### 🔍 RCA
When the server returned a new refresh token, the SDK only updated the cache for the specific `audience` and `scope` of that request. Other cache entries with different scopes but the same old refresh token were not updated, leading to failed token requests if they were used later.

### 📋 Changes
The main change is to proactively invalidate old refresh tokens from the entire cache after a successful rotation.

- Added `__tests__/Auth0Client/refresh-token-rotation.test.ts`: New tests to simulate and verify the refresh token rotation flow.
- Changed `src/Auth0Client.ts`: Implemented `_invalidateOldRefreshTokenEntries` to find and remove old refresh tokens from all relevant cache entries after a new one is received. Also improved the documentation for the `exchangeToken` method.
- Changed `package.json`: Updated dev dependencies.
- Changed `package-lock.json`: Updated dev dependencies.

### 🎯 Testing
Automated:
Unit tests were added in `__tests__/Auth0Client/refresh-token-rotation.test.ts`. These tests mock the token endpoint and verify:
- When a rotated refresh token is returned, it is correctly used in the next `getTokenSilently` call.
- The original (now invalid) refresh token is not used again.
- The SDK continues to function correctly when token rotation does not occur.

Manual:
1.  In your Auth0 Dashboard, enable "Refresh Token Rotation" for your application.
2.  Using a sample app, log in with `useRefreshTokens: true`.
3.  Call `getTokenSilently()` to get an access token.
4.  Call `getTokenSilently({ cacheMode: 'off' })` to force a token refresh. The server will return a new refresh token.
5.  Call `getTokenSilently()` again. The SDK should use the new refresh token from step 4, and the call should succeed.


src/Auth0Client.ts
```diff
@@ -1130,11 +1144,23 @@ export class Auth0Client {
...
+      if (authResult.refresh_token !== cacheEntry.refresh_token) {
+        await this._invalidateOldRefreshTokenEntries(cacheEntry.refresh_token);
+      }
+    }
 
     await this._saveEntryInCache({
       ...authResult,
       decodedToken,
-      scope: options.scope,
+      scope: scope,
```
tusharpandey13, line 1163: Make sure cache entry contains correct scope. 
This can be either an explicit case like `loginWithPopup` or a case where an existing cache entry contains a more generically-scoped RT but the current request is for a more restrictive scope. 
In that case, the generic-scoped RT itself is cached again, since SPA does NOT support token descoping.

src/Auth0Client.ts
```diff
@@ -988,9 +995,12 @@ export class Auth0Client {
       const tokenResult = await this._requestToken({
         ...options.authorizationParams,
         grant_type: 'refresh_token',
-        refresh_token: cache && cache.refresh_token,
+        refresh_token: cache?.refresh_token,
         redirect_uri,
+        cacheEntry: cache as CacheEntry,
         ...(timeout && { timeout })
+      }, {
+        scope: options.authorizationParams.scope
```
tusharpandey13, line 1003: make sure to pass current `cacheEntry` to enable fetching scope from said `cacheEntry` during re-caching step.

src/Auth0Client.ts
```diff
@@ -1150,65 +1176,81 @@ export class Auth0Client {
...
+      // Log for debugging/monitoring purposes
+      if (updatedCount > 0) {
+        console.debug(`Invalidated refresh tokens in ${updatedCount} cache entries`);
+      }
+    } catch (e) {
+      // Log the error but don't throw - this is a security cleanup operation
+      // and shouldn't break the main token refresh flow
+      console.warn('Failed to invalidate old refresh token entries:', e);
+    }
+  }
```
tusharpandey13: Update cache entry containing old, rotated RT to remove the cached RT.

nandan-bhat (dismissed):

src/cache/cache-manager.ts
```diff
@@ -143,6 +143,58 @@ export class CacheManager {
...
+    const allKeys = await this.getCacheKeys();
+    if (!allKeys) return 0;
+
+    // Filter keys to only those belonging to this client
+    const clientKeys = allKeys.filter(key => {
+      try {
+        return CacheKey.fromKey(key).clientId === clientId;
+      } catch {
+        // Fallback to string matching for keys that can't be parsed
+        return key.includes(`::${clientId}::`);
```
gyaneshgouraw-okta, line 167: Have we considered this fallback thoroughly, can there be potential side effects, as this is a more liberal condition.

gyaneshgouraw: > return key.includes(`::${clientId}::`);

Have we considered this thoroughly for potential side effects,  as this is a more liberal check.

src/cache/cache-manager.ts
```diff
@@ -143,6 +143,58 @@ export class CacheManager {
...
+      }
+    });
+
+    let updatedCount = 0;
+    for (const key of clientKeys) {
+      try {
+        const entry = await this.cache.get<WrappedCacheEntry>(key);
+
+        if (entry && ('body' in entry && 'expiresAt' in entry) && shouldUpdate(entry)) {
+          const updatedEntry = updateEntry(entry);
```
gyaneshgouraw, line 178: > const updatedEntry = updateEntry(entry);

Can you help me understand what is this line supposed to do.
Set the refresh token value in entry as `undefined` ?

Maybe we can add a line comment mentioning it

src/Auth0Client.ts
gyaneshgouraw: Small clarifiction,
So essentially we are putting together all available scopes to be preserved ?

gyaneshgouraw-okta (dismissed):

src/Auth0Client.ts
tusharpandey13: Yes, this ensures that all RTs are captured and the most generically scoped are cached from all flows.

src/cache/cache-manager.ts
tusharpandey13: Essentially, yes this allows `refresh_token` to be set as `undefined` for this specific use-case.

More generally, this acts as a map operation on the cache entries, where entries are seclected by the callback `shouldUpdate: (entry: WrappedCacheEntry) => boolean;` and mutated by the callback `updateEntry: (entry: WrappedCacheEntry) => WrappedCacheEntry | null;`

In this specific use-case, the mutation logic updates the `refresh_token` field to `undefined`.
Explicit logic to make `refresh_token` `undefined` was not put in `cache-manager.ts` to ensure Seperation-Of-Concerns is followed, `_invalidateOldRefreshTokenEntries` contains the business logic pertaining to this specific use case.

tusharpandey13: Will add comments.

src/cache/cache-manager.ts
tusharpandey13: The redundant fallback has been removed.

gyaneshgouraw-okta:
@claude

yogeshchoudhary147:
Closing this as part of our backlog cleanup (internal PRs > 3 months old).

The branch is preserved, so feel free to reopen if this becomes active again.

PR #1386, opened 2025-07-30
Add Support for Multi-Resource Refresh Tokens (MRRT)

aridibag:
### Changes

The aim of this PR, is to add the possibility to refresh tokens using MRRT functionality. In the `get` method from `cache-manager`, now we can find a RT to refresh a different audience from the ones that are already saved in the cache, letting the user change between audiences.

The user can now decide whether to use MRRT in their requests. If they want to enable it, `useMrrt` needs to be sent as an option when configuring `Auth0Client`.

To know more about MRRT and how to configure it visit: 
- [Multi-Resource Refresh Token](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token)
- [Configure and Implement MRRT](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token/configure-and-implement-multi-resource-refresh-token)

What things have been changed: 
- When `useMrrt` is enabled, on `cache-manager` `get` method, if we didn't find any entry with 100% of coincidence or any matching one, we are going to use the first `refresh_token `we find to use in the refresh request.
- When `useMrrt` is enabled, we are going to send audience and scope to the server when doing refresh. To avoid doing downscoping with that (to request less scopes than the ones in the `refresh_token` to be used), we are going to check what scopes are we going to send to the server before the refresh request using `getScopeToRequest` method. 
- To avoid leaving in the cache a `refresh_token` that has already been used (if we use it again, the refresh request will fail), after obtaining a new `refresh_token`, we will check what entries had the previous `refresh_token` and update it with the new one with `cacheManager.updateEntry`.
- After refreshing with MRRT, some scopes might not appear because they might no be inside the refresh policies. We are going to check if the refresh has been done with MRRT, and if some scopes are missing we will try to request a new token using silent auth.
- When using `cacheLocation='memory'`, we are going to save the last refresh_token obtained in a key named `latest_refresh_token`, and use that when doing refresh with MRRT (if we are not going to downscope).

### Testing

Manual:

To test it you will need to firstly follow the configuration [guide](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token/configure-and-implement-multi-resource-refresh-token), and create two APIs in the dashboard. For both cache locations (`memory`|`localstorage`), try:

1. Initialize a new Auth0 Client with `useMRRT` to `true`, and with this configuration:
```
await createAuth0Client({
  domain: '<AUTH0_DOMAIN>',
  useRefreshTokens: true,
  useRefreshTokensFallback: true,
  useMrrt: true,
  authorizationParams: {
    redirect_uri: '<MY_CALLBACK_URL>'
  }
```
2. Login to the application with your preferred login method and use `getTokenSilently` to request an access_token for the API_1.
3. After the `access_token` + `refresh_token` from API_1 has been stored, use again `getTokenSilenty` to request an `access_token` for API_2 this time. Internally it will use `refresh_token` from API_1 to do the refresh.

Automated:
Unit tests have been added in:
- `__tests__/cache/cache-manager.test.ts`. Checks that the `refresh_token` from the previous stored audience is returned for the second audience.
- `__tests__/token.worker.test.ts`. Checks that the `refresh_token` obtained from the server for the first audience, is stored as `latest_refresh_token` on worker memory, and is used to do a refresh request for the second audience.

### Checklist


src/cache/cache-manager.ts
```diff
@@ -207,4 +216,57 @@ export class CacheManager {
...
+
+      if (entry?.body?.refresh_token === oldRefreshToken) {
+        const cacheEntry = {
+          ...entry.body,
+          refresh_token: newRefreshToken,
+        } as CacheEntry;
+
+        await this.set(cacheEntry);
+      }
+    })
```
tusharpandey13: https://stackoverflow.com/a/70946414/4619444
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description:~:text=forEach()%20expects%20a%20synchronous%20function%20%E2%80%94%20it%20does%20not%20wait%20for%20promises.%20Make%20sure%20you%20are%20aware%20of%20the%20implications%20while%20using%20promises%20(or%20async%20functions)%20as%20forEach%20callbacks)

> `forEach()` expects a synchronous function — it does not wait for promises. Make sure you are aware of the implications while using promises (or async functions) as forEach callbacks

should this be a `for..of` loop maybe?

tusharpandey13:
@aridibag can you sign your commits please, this repo requires commits to be signed in order to be merged.

EXAMPLES.md
```diff
@@ -80,6 +80,28 @@ If the fallback mechanism fails, a `login_required` error will be thrown and cou
 
 **Note**: This fallback mechanism does still require access to the Auth0 session cookie, so if third-party cookies are being blocked then this fallback will not work and the user must re-authenticate in order to get a new refresh token.
 
+### Using Multi-Resource Refresh Tokens
+
+Refresh tokens from one API can be used to request new access tokens from another API. [Read more about how MRRT works for browser-based applications](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token) to help you decide wether or not you need to use this funcfionality.
```
gyaneshgouraw: Minor correction : I believe the statement should be, as in instead of from it should be for.

> Refresh tokens from one API can be used to request new access tokens _**for**_ another API.

src/cache/cache-manager.ts
aridibag: Yes, that's true!

EXAMPLES.md
aridibag: Thanks!

tusharpandey13 (dismissed):
lgtm

src/api.ts
```diff
@@ -21,7 +22,8 @@ export async function oauthToken(
   const allParams = {
     ...options,
     ...(isTokenExchange && audience && { audience }),
-    ...(isTokenExchange && scope && { scope })
+    ...(isTokenExchange && scope && { scope }),
+    ...(options.grant_type === 'refresh_token' && useMrrt && { audience, scope })
```
frederikprijck: What happens when a user has set `useMrrt` to `true`, and requests a new token for the same audience, but with different scopes? Will the token be descoped? If yes, is this expected? Isn't that considered a breaking change?

In the event where we believe this is expected, can the SDK handle Access Token properly? As this wasn't a feature that was supported or even accounted for.

src/worker/token.worker.ts
```diff
@@ -6,7 +6,9 @@ let refreshTokens: Record<string, string> = {};
 
 const cacheKey = (audience: string, scope: string) => `${audience}|${scope}`;
 
-const getRefreshToken = (audience: string, scope: string) =>
+const cacheKeyContainsAudience = (audience: string, cacheKey: string) => cacheKey.startsWith(`${audience}|`);
+
+const getRefreshToken = (audience: string, scope: string): string | undefined =>
```
aridibag, line 12: @frederikprijck I have a question about getting refresh_tokens inside the worker. If the key audience+scopes doesn't match exactly the ones stored inside the array of tokens, it never does a refresh. I'm talking to a normal refresh without MRRT. For example, if the token stored has this key `orders|read:orders create:orders`, and in the second petition we need to do a refresh with `orders|create:orders`, will this never use the refresh_token stored in the first key?

frederikprijck: We should be able to test this, but from a quick look, it looks like it wont refresh indeed.

aridibag: Isn't it weird that when using localstorage as the cache we are looking for the best match in the tokenset storage to do a refresh, but when using the worker we only search for the exact match? 🤔

src/cache/shared.ts
```diff
@@ -82,6 +82,8 @@ export type CacheEntry = {
   client_id: string;
   refresh_token?: string;
   oauthTokenScope?: string;
+  oldScopes?: string;
+  oldAudience?: string;
```
frederikprijck: Why do we need these two in the cache?

src/Auth0Client.utils.ts
```diff
@@ -96,3 +96,69 @@ export const patchOpenUrlWithOnRedirect = <
...
+/**
+ * @ignore
+ * 
+ * Checks if all scopes are included inside other array of scopes
+ */
+export const allScopesAreIncluded = (scopesToCheck?: string, scopesContainer?: string): boolean => {
+  const oldScopes = scopesContainer?.split(" ") || [];
+  const newScopes = scopesToCheck?.split(" ") || [];
+  return newScopes.every((scope) => oldScopes.includes(scope));
+}
```
frederikprijck: I am confused about naming here. What is a scopes container, what are old scopes and what are new scopes?

src/Auth0Client.utils.ts
```diff
@@ -96,3 +96,69 @@ export const patchOpenUrlWithOnRedirect = <
...
+ * @param useMrrt Setting that the user can activate to use MRRT in their requests
+ * @param authorizationParams Contains the audience and scope that the user requested to obtain a token
+ * @param oldAudience Audience stored with the refresh_token wich we are going to use in the request
+ * @param oldScope Scope stored with the refresh_token wich we are going to use in the request
+ */
+export const getScopeToRequest = (
+  useMrrt: boolean | undefined,
+  authorizationParams: { audience?: string, scope: string },
+  oldAudience?: string,
+  oldScope?: string
```
frederikprijck: I think old is a weird prefix here, I am not sure what it means. Why is it old? Isnt this the audience and scope from the cacheEntry?

src/api.ts
```diff
@@ -20,10 +21,14 @@ export async function oauthToken(
   const isTokenExchange =
     options.grant_type === 'urn:ietf:params:oauth:grant-type:token-exchange';
 
+  const refreshWithMrrt =
+    options.grant_type === 'refresh_token' && useMrrt;
```
frederikprijck, line 25: So when the grant_type is refresh_token and useMrrt is true, we pass through scope and audience. How does this solve the issue with avoiding descoping?

src/cache/cache-manager.ts
```diff
@@ -207,4 +225,57 @@ export class CacheManager {
...
+
+  /**
+   * Returns the first entry that contains a refresh_token that satisfies the following conditions
+   * The keys inside the cache are in the format {prefix}::{clientId}::{audience}::{scope}.
+   * - `prefix` is strict equal to Auth0's internally configured `keyPrefix`
+   * - `clientId` is strict equal to the `cacheKey.clientId`
+   * @param keyToMatch The provided cache key
+   * @param allKeys A list of existing cache keys
+   */
+  private async getEntryWithRefreshToken(keyToMatch: CacheKey, allKeys: Array<string>): Promise<Partial<CacheEntry> | undefined> {
```
frederikprijck, line 241: This just give any refresh token, how do we know this is the correct one?

frederikprijck (changes requested):
In general, I would like to see a bit more documentation on why certain things are done, both to better understand what is happening as well as to document for future reasons and maintanance. I also think we should resolve a few namings to ensure our names our also more self explanatory, things like oldAudience mean nothing without more context.

src/cache/shared.ts
aridibag: These two properties are not saved in the cache. If you see the [modifiedCachedEntry](https://github.com/auth0/auth0-spa-js/pull/1386/files#diff-640190c0d5c04d907bd08f38a8f4b8d53140b1c63acb04185e5e86c250302ac0R130-R131) method in cache-manager, you will see that we are only returning them when doing `this.cacheManager.get`. But, instead of using that naming, I decided to change them to `audience` and `scope` again.

src/Auth0Client.utils.ts
aridibag: Yes, it is a bit confusing 😅 
To name them in a more generic way, I changed them to `scopesGroupA` and `scopesGroupB`. Because I only want to see if a group is contained inside the other. What do you think?

src/Auth0Client.utils.ts
aridibag: Changed it in a comment above!

src/cache/shared.ts
frederikprijck: I think this is confusing and I am not sure I understand what their purpose is and what we have them. It would be helpful to make the intent more clear.

src/Auth0Client.utils.ts
frederikprijck: So i think a function named `allScopesAreIncluded()` could take in: scope: string, scopesToInclude: string[])

```ts
export const allScopesAreIncluded = (scope?: string, scopesToInclude?: string[]): boolean => {
  const scopes = scope.split(" ") || [];
  return scopesToInclude.every((scope) => scopes.includes(scope));
}
```

src/api.ts
aridibag: We are going to always send audience + scope to the server while useMrrt=true, but now before doing the refresh request, we are using [getScopeToRequest](https://github.com/auth0/auth0-spa-js/pull/1386/files#diff-ef29a8a5329e7e9e0d25dca29842c63ee35ab5c04b61c3c39fbece134cc77886R1014) method. With that we are checking if we are going to do downscoping checking the audience and scopes requested by the user. If we see that we are going to request the same scopes as the refresh_token to use, but with some missing, we are going to incorporate that missing scope to the ones we are going to request. So with that, will will avoid the downscoping.

src/cache/cache-manager.ts
aridibag: This is the final step when searching for a refresh token. If we didn't find any entry with 100% matching coincidence or at least to contain all requested scopes in the method, we are going to use any refresh_token we have available in the cache. When we are doing MRRT, as we specify what audience and scopes we want to the server, it can be used any refresh_token to do the request.

src/api.ts
frederikprijck: So if useMrrt is false, we wont sent scopes. If useMrrt is true, we sent scopes? Even when MRRT is not involved? And we just rely on the fact that we didnt make a mistake with the scopes?

I think it'd be safer to not change existing behavior. But leaving it to you, if you are confident this will not cause issues.

frederikprijck: Personallyy I think new behavior for MRRT should only kick in when the audience is different and useMrrt is set to true.

src/cache/cache-manager.ts
frederikprijck: But how do we know _any_ works? Today, customers can have multiple RT's in their store, for different audiences. If we release this change, how are we going to know which RT to use? Afaik we need to have a RT for the audience that is part of the MRRT policy, no ?

src/cache/shared.ts
aridibag: Already removed them :)

src/Auth0Client.utils.ts
aridibag: I modified the method a little bit with your suggestion, but the `scopesToInclude` are still a string to transform them into an array inside here. If not I should do a `scope.split(" ") || [];` outside the method each time I want to use it.

mogzol:
Noticed some more strange behaviour:

Lets say I am working with two access tokens, one for the default API, and one for API 2 (with audience "https://example.com")

Initially after logging in, the cache will just have a token for just the default API, this makes sense so far.

Now, if I call `getTokenSilently` to get a token using MRRT for API 2, using `cacheMode: "cache-only"`, then, as expected, no access token is returned. But, a cache entry is still created for the "https://example.com" token. This cache entry only has the refresh token in it, no access token (which makes sense). This seems a little weird but is still fine.

Next, if I refresh the default access token (forcing a request using `cacheMode: "off"`), then the cache entry for the default access token is updated correctly, but the cache entry for API 2 is **not** updated, and so still has an old refresh token. Additionally, a new cache entry is created for the default token but with no client ID in the cache key. This is weird and seems broken.

Since the API 2 cache entry was not updated, if I now try to call `getTokenSilently` for API 2, the request fails because it uses the old, now invalid refresh token.

Also, even after logging out, that extra cache entry for the default access token is still in the cache, I guess because it doesn't have the client ID in the key, so it doesn't get cleared.

Sorry if that is hard to follow, here is an example reproduction. You'll have to update the domain, client ID, and audience to be valid for your environment, and then make sure your localStorage is empty before running it:
<details>
  <summary>Reproduction Code</summary>

```js
const API_2_AUTH_PARAMS = { audience: "https://secondapi.mysite.com" }

const auth0Client = new Auth0Client({
  domain: "mytenant.auth0.com",
  clientId: "MyClientId",
  useRefreshTokens: true,
  useMrrt: true,
  cacheLocation: "localstorage",
  authorizationParams: {
    redirect_uri: window.location.origin,
  },
});

run();

async function run() {
  // Check if we're logged in by trying to get the default token
  try {
    await auth0Client.getTokenSilently()
  } catch {
    console.log("Not logged in")
    if (window.location.search.includes("code=")) {
      await auth0Client.handleRedirectCallback();
      window.history.replaceState({}, document.title, window.location.pathname);
      console.log("Handled login callback");
    } else {
      console.log("Redirecting to login...");
      await auth0Client.loginWithRedirect();
      return;
    }
  }

  // Get cached token. If the token doesn't exist, this still adds an entry to the cache for it.
  const cachedTokenApi2 = await auth0Client.getTokenSilently({
    cacheMode: "cache-only",
    authorizationParams: API_2_AUTH_PARAMS,
  });

  console.log("cachedTokenApi2", cachedTokenApi2);

  // Now refresh the default access token
  const newDefaultToken = await auth0Client.getTokenSilently({ cacheMode: "off" });

  console.log("newDefaultToken", newDefaultToken);

  // At this point, two weird things have happened.
  //  - First, the refresh token in the API 1 cache entry has NOT been updated.
  //  - Second, an extra cache entry has been made for the default token which DOES NOT have
  //    the client ID in the cache key. This cache entry will persist even after calling logout()

  // If we now try to get a token for API 1 again, this time not using cache-only, it will make the
  // request using the outdated refresh token and fail
  const newTokenApi2 = await auth0Client.getTokenSilently({
    cacheMode: "off",
    authorizationParams: API_2_AUTH_PARAMS,
  });

  console.log("newTokenApi2", newTokenApi2);
}
```
</details>

That should initially redirect you to login, and then once logged in, recreate the issue. If you look at the console logs once it is done, you should see an error "Unknown or invalid refresh token" from when it tried to get a new token for API 2. If you look at your local storage at this point, you should see the issue I described above.

mogzol:
Noticed something else, not sure if this is intentional or not, but all of the scopes specified in the initial `authorizationParams` are included in MRRT requests. So if I initialize Auth0Client like:

```js
const auth0Client = new Auth0Client({
  authorizationParams: {
    audience: "https://default.com",
    scope: "default:permission:a default:permission:b"
  },
  ...
});
```

And then later try to call refreshTokenSilently using MRRT for a different API like:

```js
auth0Client.getTokenSilently({
  authorizationParams: {
    audience: "https://other.com",
    scope: "other:permission:a other:permission:b"
  },
});
```

Then the scopes included in that request will have both the scopes from the default API and the new ones: `default:permission:a default:permission:b other:permission:a other:permission:b` which seems wrong to me? I would expect it to only request the scopes included in the getTokenSilently request.

mogzol:
One more thing I found with scopes, if I request scopes which are not returned in the response, like this:

```js
auth0Client.getTokenSilently({
  authorizationParams: {
    audience: "https://example.com",
    scope: "real:permission something:else"
  },
});
```

Where the response only includes `real:permission`, then the `getTokenSilently` call throws an error:

```
Missing Refresh Token (audience: 'https://example.com', scope: 'openid profile email offline_access real:permission something:else')
```

But the returned token is still added to the cache. So if I make the exact same call again:

```js
auth0Client.getTokenSilently({
  authorizationParams: {
    audience: "https://example.com",
    scope: "real:permission something:else"
  },
});
```

Then this time it returns the cached token and does not throw an error.

tusharpandey13:
> Noticed something else, not sure if this is intentional or not, but all of the scopes specified in the initial `authorizationParams` are included in MRRT requests. 
Is that how it is supposed to work? I would expect it to only request the scopes included in the getTokenSilently request.

Hi @mogzol please take a look at https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token
> When using MRRT, Auth0 merges two sources of authorization to determine what access token to issue during a refresh token exchange:
- The audience and scopes granted in the original authentication flow.
- The audience and scopes configured in the application’s MRRT policy.

src/cache/cache-manager.ts
```diff
@@ -114,6 +118,24 @@ export class CacheManager {
...
+  private async modifiedCachedEntry(wrappedEntry: WrappedCacheEntry, cacheKey: CacheKey): Promise<Partial<CacheEntry>> {
+    // We need to keep audience and scope in order to check them later when doing refresh
+    // using MRRT. See getScopeToRequest method.
+    wrappedEntry.body = {
+      refresh_token: wrappedEntry.body.refresh_token,
+      audience: wrappedEntry.body.audience,
+      scope: wrappedEntry.body.scope,
+    };
+
+    await this.cache.set(cacheKey.toKey(), wrappedEntry);
```
tusharpandey13, line 130: > `await this.cache.set(cacheKey.toKey(), wrappedEntry);`

Maybe this should not be done if:
1. `cache-only` is being used.
2. `cacheKey.clientId` is `undefined`.

This will prevent the phantom cache entry problem when `cache-only` is used and an AT is requested in MRRT flow for a different audience than the currently cached one. This entry cannot be updated when a RT rotation happens.

aridibag: The method `modifiedCachedEntry`, which has part of the code you selected, is being used as well to return a refresh_token inside the cache for a regular refresh, [as it has been working nowadays](https://github.com/auth0/auth0-spa-js/pull/1386/files#diff-640190c0d5c04d907bd08f38a8f4b8d53140b1c63acb04185e5e86c250302ac0L104). That method is being used to return a refresh_token for a MRRT refresh as well. That's why, to fix the phantom entry, [I only added ](https://github.com/auth0/auth0-spa-js/pull/1386/files#diff-640190c0d5c04d907bd08f38a8f4b8d53140b1c63acb04185e5e86c250302ac0R94)the `cacheMode !== 'cache-only'` condition in the specific part of MRRT, to avoid changing nowaday's behaviour.

src/Auth0Client.utils.ts
gyaneshgouraw: I think `cachedAudience` and `cachedScope` might be clearer alternatives to the current names prefixed with old.

PR #1388, opened 2025-07-30
Add support for DPoP

martinml:
### Changes

> [!IMPORTANT]
> This PR is only the first part of the whole DPoP feature. The second PR is #1400.
> 
> Because of this, **please do not make a new release until both PRs are merged**.

This PR adds support for DPoP. It includes:

- Key generation and secure storage.
- Authorization code binding (i.e. `dpop_jkt` parameter).
- Access token request binding (i.e. `DPoP` header).
- Automatic nonce management and retries for Auth0 flows.

It does NOT include:
- Docs for the DPoP feature.
- A way of using DPoP with external, custom APIs.

Both of these ☝️ will be added in #1400, for an easier review.

ℹ️ Notes to understand some changes:

- Since some changes to core code were made (i.e. header parsing and `token_type` property in cache), some apparently unrelated tests had to be adapted.
- Since the `dpop` package has some syntax that is too new for the current Jest configuration, transpilation had to be explicitely enabled for it:
    - The `ts-jest` preset was changed to `ts-jest/presets/js-with-ts`.
    - `allowJs` was enabled.
    - `dpop` was added to `transformIgnorePatterns`.
- I added `fake-indexeddb` for testing, which needs a `structuredClone` implementation in `jsdom` but current version doesn't include it, so we use the Node implementation instead.
- `ES2019.Object`, `DOM` and `DOM.Iterable` were added to the TS `lib`s so Fetch API's `Headers` interface can be used. This will not change the resulting code, just the types available when type-checking.

### Testing

### Checklist


__tests__/dpop/storage.test.ts
```diff
@@ -0,0 +1,241 @@
...
+      it('does not create handle', () =>
+        expect(storage['createDbHandle']).not.toHaveBeenCalled());
+    });
+
+    describe('otherwise', () => {
+      beforeEach(() => {
+        storage['dbHandle'] = undefined;
+      });
+
+      beforeEach(() => storage['getDbHandle']());
```
Widcket, line 66: Are multiple `beforeEach` calls needed? Seems to be a common pattern across the tests.

Widcket: Also seems like the "act" part of arrange-act-assert is being done in a `beforeEach`.

martinml: > Are multiple `beforeEach` calls needed?

Not really needed, no; it's just a convention for organization: the first `beforeEach` is for the "arrange" part and the second one for the "act". They could be merged together, the result is identical.

> Also seems like the "act" part of arrange-act-assert is being done in a `beforeEach`.

Correct! I find that this pattern naturally leads to writing tests that don't interfere with each other (of which there are several in `auth0-spa-js` 😓).

Also, when you need to do more than one `it()`, you don't need to repeat the "act" part for each one.

Alternatively, if you were to put the "act" directly inside a `describe()`:
1. Unexpected exceptions would be difficult to track down due to how Jest reports them. They would also "silence" any other successful tests.
2. It would be executed immediately upon loading the suite, which would make using `describe.only()`/`describe.skip()` more difficult.

src/dpop/dpop.ts
```diff
@@ -0,0 +1,56 @@
...
+  public setNonce(nonce: string, id?: string): Promise<void> {
+    return this.storage.setNonce(nonce, id);
+  }
+
+  protected async getOrGenerateKeyPair(): Promise<dpopUtils.KeyPair> {
+    let keyPair = await this.storage.findKeyPair();
+
+    if (!keyPair) {
+      keyPair = await dpopUtils.generateKeyPair();
+      this.storage.setKeyPair(keyPair);
```
Widcket: Doesn't this need to be awaited?

martinml: Agh 100% it does, good catch! Fixed.

src/utils.ts
```diff
@@ -246,3 +246,18 @@ export const parseNumber = (value: any): number | undefined => {
   }
   return parseInt(value, 10) || undefined;
 };
+
+/**
+ * Ponyfill for `Object.fromEntries()`, which is not available until ES2020.
```
Widcket, line 251:
```suggestion
 * Polyfill for `Object.fromEntries()`, which is not available until ES2020.
```

src/Auth0Client.ts
```diff
@@ -1211,6 +1230,64 @@ export class Auth0Client {
...
+   *              select a specific non-Auth0 nonce.
+   */
+  public getDpopNonce(id?: string): Promise<string | undefined> {
+    this._assertDpop(this.dpop);
+
+    return this.dpop.getNonce(id);
+  }
+
+  /**
+   * Gets the current DPoP nonce used for making requests to Auth0.
```
Widcket:
```suggestion
   * Sets the current DPoP nonce used for making requests to Auth0.
```

src/dpop/storage.ts
```diff
@@ -0,0 +1,134 @@
...
+  public setKeyPair(keyPair: KeyPair): Promise<void> {
+    return this.save(TABLES.KEYPAIR, this.buildKey(), keyPair);
+  }
+
+  protected async save(
+    table: Table,
+    key: IDBValidKey,
+    obj: unknown
+  ): Promise<void> {
+    await this.executeDbRequest(table, 'readwrite', table =>
```
Widcket: Since this method returns a promise, does it need the `await` (and thus be marked as `async`)?

src/utils.ts
martinml: I also find the naming weird, but [it is apparently correct](https://ponyfoo.com/articles/polyfills-or-ponyfills) 😄

src/dpop/storage.ts
martinml: The code is correct although it definitely falls a bit short in showing the intention.

The idea here is that I want `save()` to return `Promise<void>` but `executeDbRequest()` returns `Promise<IDBValidKey>`, so I have to "discard" that return type somehow. Otherwise: `Type 'IDBValidKey' is not assignable to type 'void'.`

So one option is doing what I did. Another would be removing the `async` keyword and then doing `return this.executeDbRequest(...).then(() => {})`, but that feels awkward.

A different approach would be keeping the `async` keyword but then `void` the result: `return void this.executeDbRequest(...)`, which might show the original intention a little better.

Or I can just add a comment 😄 

Any preference from those?

src/Auth0Client.ts
martinml: Fixed!

src/dpop/storage.ts
Widcket: > A different approach would be keeping the async keyword but then void the result: return void this.executeDbRequest(...), which might show the original intention a little better.

Agreed.

martinml: Fixed!

Widcket:
@martinml CI is failing

Widcket (approved):

PR #1442, opened 2025-10-16
[ESD-52771] Support mix of DPoP and Bearer in the fetcher

adamjmcgrath:
### Changes

Applications can have a mix of APIs that use Bearer or DPoP, so the fetcher needs to be able to make this decision based on the access token type.

This is added in a non breaking way, when `getAccessToken` returns a `string` - the behaviour remains the same (the token type is configured by the `dpopNonceId` option)

### References

https://auth0team.atlassian.net/browse/ESD-52771

### Testing

- [ ] This change has been tested on the latest version of the platform/language

### Checklist

- [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [ ] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
- [ ] All code quality tools/guidelines have been run/followed

### DIFF
--- a/__tests__/fetcher.test.ts
+++ b/__tests__/fetcher.test.ts
@@ -648,12 +648,26 @@ describe('Fetcher', () => {
     });
 
     describe('when DPoP is enabled', () => {
-      const getAccessTokenMock = jest.fn().mockResolvedValue({ access_token: TEST_ACCESS_TOKEN, token_type: 'Bearer' });
+      const getAccessTokenMock = jest.fn().mockResolvedValue({
+        access_token: TEST_ACCESS_TOKEN,
+        token_type: 'Bearer'
+      });
       const fetchStub = jest.fn().mockResolvedValue(new Response());
-      const fetcher = newTestFetcher({
-        getAccessToken: getAccessTokenMock,
-        dpopNonceId: 'foo', // Added DpoP nonce ID to enable DPoP, but still uses Bearer token
-        fetch: fetchStub
+      const fetcher = newTestFetcher(
+        {
+          getAccessToken: getAccessTokenMock,
+          dpopNonceId: 'foo', // Added DpoP nonce ID to enable DPoP, but still uses Bearer token
+          fetch: fetchStub
+        },
+        {
+          getDpopNonce: jest.fn().mockResolvedValue(TEST_DPOP_NONCE),
+          generateDpopProof: jest.fn().mockResolvedValue(TEST_DPOP_PROOF)
+        }
+      );
+
+      beforeEach(() => {
+        fetchStub.mockClear();
+        getAccessTokenMock.mockClear();
       });
 
       it('should honor bearer token type', async () => {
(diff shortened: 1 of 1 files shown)


src/fetcher.ts
```diff
@@ -133,11 +139,12 @@ export class Fetcher<TOutput extends CustomFetchMinimalOutput> {
...
   protected setAuthorizationHeader(
     request: Request,
-    accessToken: string
+    accessToken: string,
+    tokenType: string = TokenType.Bearer
   ): void {
     request.headers.set(
       'authorization',
-      `${this.config.dpopNonceId ? 'DPoP' : 'Bearer'} ${accessToken}`
+      `${tokenType} ${accessToken}`
```
gyaneshgouraw, line 147: Hey @adamjmcgrath, I’m trying to understand this change a bit better.

In which scenario would the check fail for this string:
${this.config.dpopNonceId ? 'DPoP' : 'Bearer'} ${accessToken}?

From what I gather, if we have a bearer token, the expectation was that `dpopNonceId` would be absent. Even so, the original check would have still worked in that case.

I do see the intention behind the new approach using ${tokenType} ${accessToken}, it’s definitely more accurate and deterministic by explicitly relying on the token type.

Wanted to gather if i was understanding it correct.

src/Auth0Client.ts
```diff
@@ -1527,14 +1523,6 @@ export class Auth0Client {
   public async connectAccountWithRedirect<TAppState = any>(
     options: RedirectConnectAccountOptions<TAppState>
   ) {
-    if (!this.options.useDpop) {
-      throw new Error('`useDpop` option must be enabled before using connectAccountWithRedirect.');
```
gyaneshgouraw, line 1531: I believe, with this change the previous PR might not be required  
https://github.com/auth0/auth0-spa-js/pull/1441

adamjmcgrath: Have closed that in favour of this PR for simplicity

src/fetcher.ts
adamjmcgrath: > In which scenario would the check fail for this string:
> ${this.config.dpopNonceId ? 'DPoP' : 'Bearer'} ${accessToken}?

If you configure the fetcher to use DPoP but the Authorization Server is still issuing Bearer tokens. This was found when someone was testing without the dpop flag enabled on their tenant. But a real world example would be you'd need it when migrating from Bearer to DPoP

src/Auth0Client.ts
```diff
@@ -1527,14 +1523,6 @@ export class Auth0Client {
...
-    }
-
-    if (!this.options.useMrrt) {
-      throw new Error('`useMrrt` option must be enabled before using connectAccountWithRedirect.');
```
gyaneshgouraw, line 1535: Hey @adamjmcgrath, just to confirm, do we not need MRRt as a mandatory feature for this?

gyaneshgouraw: Ok, users can use silent auth/web_message also, so MRRT would not be mandatory.

PR #1456, opened 2025-11-06
fix: clear cache when different user logs in without logout

### LINKED ISSUE #1433: Library can return tokens for different users depending on the audience in the request if users log in without first calling `logout`
### Checklist

- [x] The issue can be reproduced in the [auth0-spa-js sample app](https://github.com/auth0-samples/auth0-javascript-samples/tree/master/01-Login) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/auth0-spa-js#readme), [Examples](https://github.com/auth0/auth0-spa-js/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-spa-js/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://auth0.com/docs/libraries/auth0-single-page-app-sdk) and [API documentation](https://auth0.github.io/auth0-spa-js/), and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-spa-js/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I am not sure if this is actually an issue, but it seems like very strange behaviour to me. If this behaviour is intended then this can be closed.

The issue is that when using refresh tokens, if `logout` is not called before logging in, it is possible to get into a state where depending on what audience you use in the `getTokenSilently` request, the token returned will be for completely different users. There is no indication that this is happening, and the ID token in the cache will change to be whatever user the most recently updated token was for. Additionally, it's not even possible to check what user the returned access token is for without decoding it, and this library doesn't provide any way to do that.

The reason this happens is that logging in only updates the default token in the cache, and does not update or clear any other cached tokens.

This seems especially likely to happen when using MRRT, since there are likely to be multiple tokens in the cache for different audiences. But it can also happen without MRRT if `useRefreshTokensFallback` is enabled.

### Reproduction

1. Create a new Auth0Client instance with `useRefreshTokens: true` and either `useMrrt` or `useRefreshTokensFallback` set to `true`, to allow us to request tokens for multiple audiences.
2. Log in using either `loginWithPopup` or `loginWithRedirect`
3. Request a token **for a new audience**:
   ```js
   auth0Client.getTokenSilently({ authorizationParams: { audience: "https://second-api.example.com" }})
   ```
4. Without clearing the cache (so without calling `logout`), log in again with `loginWithPopup` or `loginWithRedirect` **as a different user**.
5. Request the token for the second audience again:
   ```js
   auth0Client.getTokenSilently({ authorizationParams: { audience: "https://second-api.example.com" }})
   ```
6. Observe that the returned token is a token for the first user, and **not the newly logged in user**. If you instead call  `auth0Client.getTokenSilently()` without any parameters to get the default token, then that token **is** for the newly logged in user.
7. Also observe that even if you use `{ cacheMode: "off" }` to force a new token to be retrieved, it will still be for the old user, since the refresh token is stored along with the cache entry.
8. If you call `getUser`, observe that the user changes depending on which token most recently updated.

### Additional context

Like I mentioned, this issue is caused by the fact that logging in only updates the default cache entry, and not any other cached tokens.

You could argue this behaviour is intentional since `logout` was never called, but I think it is very strange to allow this to happen at all, especially since there's no easy way to detect it as we can't easily see what user an access token is for. It seems like it could only lead to problems, I can't imagine wanting this behaviour.

The simplest fix would be to just clear the cache whenever a new user logs in, that is the behaviour I would expect.

### auth0-spa-js version

2.5.0

### Which framework are you using (React, Angular, Vue...)?

n/a

### Framework version

n/a

### Which browsers have you tested in?

Chrome, Firefox, Safari

yogeshchoudhary147:
## Problem

When using refresh tokens with multiple audiences, logging in as a different user without calling `logout()` first causes **access tokens** for secondary audiences to still belong to the previous user.

**Reproduction:**
1. Login as User A
2. Request token for a secondary audience → **User A's access token gets cached**
3. Login as User B without calling `logout()`
4. Request token for the same secondary audience → Returns **User A's cached access token** ❌

**Root Cause:**
When a new user logs in, only the default access token is updated in the cache. **Access tokens for secondary audiences from the previous user remain cached**. When requesting tokens for those audiences, the SDK returns the stale cached access token or uses the cached refresh token to fetch a new token—both belonging to the previous user.

## Changes

This PR adds user switching detection during the authorization code flow. When a different user is detected (by comparing the `sub` claim), **all cached tokens** (access tokens, refresh tokens, and ID tokens) are cleared before saving the new user's tokens.

**Modified Files:**
- **src/Auth0Client.ts** - Added user switching detection in `_requestToken()`
- **__tests__/Auth0Client/handleRedirectCallback.test.ts** - Added test coverage
- **__tests__/Auth0Client/helpers.ts** - Updated mock to include `sub` in claims

**Test Coverage:**
- ✅ Cache is cleared when a different user logs in via redirect
- ✅ Cache is NOT cleared when the same user logs in again

## Impact

✅ **No breaking changes** - Behavior only affects authentication flows  
✅ **No API changes** - All public APIs remain unchanged  
✅ **Prevents token leakage** - Access tokens from different users cannot coexist in cache  
✅ **Works with MRRT** - Compatible with Multi-Resource Refresh Tokens  
✅ **Works with `useRefreshTokensFallback`** - Compatible with fallback mechanism  

## Testing

- ✅ All unit tests pass
- ✅ Manual testing with real Auth0 application
- ✅ Verified user switching scenario with multiple audiences
- ✅ Confirmed cache is properly cleared between different users

Fixes #1433

### DIFF
--- a/__tests__/Auth0Client/handleRedirectCallback.test.ts
+++ b/__tests__/Auth0Client/handleRedirectCallback.test.ts
@@ -622,4 +622,58 @@ describe('Auth0Client', () => {
       expect((client as any).transactionManager.remove).toHaveBeenCalled();
     });
   });
+
+  describe('handleRedirectCallback - user switching detection', () => {
+    it('should clear cache when different user logs in via redirect', async () => {
+      const auth0 = setup();
+
+      jest.spyOn(auth0['cacheManager'], 'clear');
+      jest.spyOn(auth0['userCache'], 'remove');
+
+      // First login as default user
+      await loginWithRedirect(auth0);
+
+      const cachedUser1 = await auth0.getUser();
+      expect(cachedUser1.sub).toBe('me');
+
+      // Configure mockVerify to return a different user for second login
+      mockVerify.mockReturnValueOnce({
+        claims: {
+          sub: 'user2',
+          name: 'User 2',
+          exp: Date.now() / 1000 + 86400
+        },
+        user: {
+          sub: 'user2',
+          name: 'User 2'
+        }
+      });
+
+      // Second login as different user
+      await loginWithRedirect(auth0);
+
+      // Verify cache was cleared
+      expect(auth0['cacheManager'].clear).toHaveBeenCalledWith(TEST_CLIENT_ID);
+      expect(auth0['userCache'].remove).toHaveBeenCalledWith('@@user@@');
+
+      // Verify new user is cached
+      const cachedUser2 = await auth0.getUser();
+      expect(cachedUser2.sub).toBe('user2');
+    });
+
+    it('should not clear cache when same user logs in again', async () => {
+      const auth0 = setup();
+
+      jest.spyOn(auth0['cacheManager'], 'clear');
+
+      // First login
+      await loginWithRedirect(auth0);
+
+      // Second login with same user
+      await loginWithRedirect(auth0);
+
+      // Verify cache was NOT cleared
+      expect(auth0['cacheManager'].clear).not.toHaveBeenCalled();
+    });
+  });
 });
--- a/__tests__/Auth0Client/helpers.ts
+++ b/__tests__/Auth0Client/helpers.ts
@@ -137,6 +137,7 @@ export const setupFn = (mockVerify: jest.Mock) => {
     mockVerify.mockReturnValue({
       claims: Object.assign(
         {
+          sub: 'me',
           exp: Date.now() / 1000 + 86400
         },
         claims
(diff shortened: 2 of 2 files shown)


gyaneshgouraw-okta:
@claude

PR #1458, opened 2025-11-10
feat: add support `scopes` parameter for connected accounts

guabu:
### 📋 Changes

Adds support for a top-level `scopes` parameter when initiating a connect account flow.

### 🎯 Testing

See included tests.

### Checklist

- [X] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [X] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
- [X] All code quality tools/guidelines have been run/followed

### DIFF
--- a/__tests__/Auth0Client/connectAccountWithRedirect.test.ts
+++ b/__tests__/Auth0Client/connectAccountWithRedirect.test.ts
@@ -50,6 +50,9 @@ describe('Auth0Client', () => {
       const options: RedirectConnectAccountOptions<any> = {
         connection: 'google-oauth2',
         scopes: ['profile', 'email'],
+        authorization_params: {
+          prompt: 'consent',
+        }
       };
 
       await client.connectAccountWithRedirect(options);
@@ -61,6 +64,9 @@ describe('Auth0Client', () => {
           state: expect.any(String),
           code_challenge: expect.any(String),
           code_challenge_method: 'S256',
+          authorization_params: {
+            prompt: 'consent',
+          }
         })
       );
     });
--- a/src/global.ts
+++ b/src/global.ts
@@ -557,6 +557,12 @@ export interface RedirectConnectAccountOptions<TAppState = any> {
 
   /**
    * Array of scopes to request from the Identity Provider during the connect account flow.
+   * 
+   * @example
+   * await auth0.connectAccountWithRedirect({
+   *   connection: 'google-oauth2',
+   *   scopes: ['https://www.googleapis.com/auth/calendar']
+   * });
    */
   scopes?: string[];
 
@@ -565,17 +571,7 @@ export interface RedirectConnectAccountOptions<TAppState = any> {
    *
    * @example
    * await auth0.connectAccountWithRedirect({
-   *   connection: 'google-oauth2',
-   *   scopes: ['https://www.googleapis.com/auth/calendar'],
-   *   authorization_params: {
-   *     // additional authorization params to forward to the authorization server
-   *   }
-   * });
-   *
-   * @example
-   * await auth0.connectAccountWithRedirect({
    *   connection: 'github',
-   *   scopes: ['repo', 'user'],
    *   authorization_params: {
    *     audience: 'https://api.github.com'
    *   }


src/MyAccountApiClient.ts
```diff
@@ -4,12 +4,12 @@ import { Fetcher } from './fetcher';
...
   /** The name of the connection to link the account with (e.g., 'google-oauth2', 'facebook'). */
   connection: string;
+  /** Array of scopes to request from the Identity Provider during the connect account flow. */
+  scopes?: string[];
   /** The URI to redirect to after the connection process completes. */
   redirect_uri: string;
   /** An opaque value used to maintain state between the request and callback. */
   state?: string;
-  /** A string value used to associate a Client session with an ID Token, and to mitigate replay attacks. */
-  nonce?: string;
```
guabu, line 12: Removed the `nonce` parameter here since it's not relevant to the Connect Account flow — it appears to be an older artifact lingering around.

__tests__/Auth0Client/connectAccountWithRedirect.test.ts
```diff
@@ -49,15 +49,15 @@ describe('Auth0Client', () => {
     it('should call myAccountApi.connectAccount with correct params', async () => {
       const options: RedirectConnectAccountOptions<any> = {
         connection: 'google-oauth2',
-        authorization_params: { scope: 'profile email' }
+        scopes: ['profile', 'email'],
```
pmalouin, line 52: Minor: perhaps this test should continue to assert correct handling of `authorization_params` (as well as `scopes`).

src/global.ts
```diff
@@ -555,23 +555,28 @@ export interface RedirectConnectAccountOptions<TAppState = any> {
...
+   */
+  scopes?: string[];
+
   /**
    * Additional authorization parameters for the request.
    *
    * @example
    * await auth0.connectAccountWithRedirect({
    *   connection: 'google-oauth2',
+   *   scopes: ['https://www.googleapis.com/auth/calendar'],
```
pmalouin: Minor: This `@example` block showcasing usage of `scopes` should be moved to the jsdoc block of the new `scopes` field above.

In this block, we can continue to show usage of `authorization_params` specifically and the second @example block below should be enough (with usage of `audience` with Github)

__tests__/Auth0Client/connectAccountWithRedirect.test.ts
guabu: Done as part of https://github.com/auth0/auth0-spa-js/pull/1458/commits/be22f7413b9de552da04b0266e2c9e564fe478c1.

src/global.ts
guabu: Done as part of https://github.com/auth0/auth0-spa-js/pull/1458/commits/be22f7413b9de552da04b0266e2c9e564fe478c1.

__tests__/Auth0Client/connectAccountWithRedirect.test.ts
gyaneshgouraw: @guabu Could we add a test case to validate how the function behaves when this payload is passed?
Since the current type definition allows this shape, we should confirm the expected behavior:
```ts
const options: RedirectConnectAccountOptions<any> = {
  connection: 'google-oauth2',
  authorization_params: { scope: 'profile email' }
};
```

gyaneshgouraw-okta:
@guabu Can we make the PR description more descriptive to include required context about this PR?
It would be helpful to include:

- The motivation for moving `scope` out of `authorizationParam`s and promoting it to a top-level field.
- A note that the new `scopes` field is now an `array of strings`, which is a meaningful API change.
- Clarification on usage: Should users still pass `scope` inside `authorizationParams`? The type definition still allows it, so it would be good to explain how that scope differs from the new top-level scopes, and which one should be preferred going forward.

This detail will help us understand more about the context of the PR

EXAMPLES.md
```diff
@@ -662,8 +662,9 @@ Use the new `connectAccountWithRedirect` method to redirect the user to the thir
 // Start the connect flow by redirecting to the thrid party API's login, defined as an Auth0 connection
 await auth0.connectAccountWithRedirect({
   connection: '<CONNECTION eg, google-apps-connection>',
-  authorization_params: {
-    scope: '<SCOPE eg https://www.googleapis.com/auth/calendar.acls.readonly>'
+  scopes: ['<SCOPE eg https://www.googleapis.com/auth/calendar.acls.readonly>'],
+  authorizationParams: {
```
gyaneshgouraw, line 666: The property key needs to be corrected. It is `authorization_params` in the type-def currently.

Also what happens, if user provides both types of scopes, as supported in type def.
We should have it clarified, if it supported or user is expected to only provide the top level scope.
```ts
scopes: ['<SCOPE eg https://www.googleapis.com/auth/calendar.acls.readonly>'],
  authorizationParams: {
      scope:"<relevant scope>"
  }
```

sam-muncke:
@gyaneshgouraw-okta I cant edit the PR description @guabu is no longer working on this. I can answer your questions though:

> The motivation for moving scope out of authorizationParams and promoting it to a top-level field.


One of the issues with the currently implementation is that it doesn't give a good way for a developer to override the default connection scopes whilst also ensure at least the minimum required scopes (typically something like offline_access and something like openid/profile/email depending on the provider) are included in the request. There has been quite a lot of thought and debate around this and it is outlined [here](https://oktawiki.atlassian.net/wiki/spaces/~724630722/pages/4020568638/Connected+Accounts+Scope+problem+Proposal+3) but suffice to say the decision has been made and the structure in this PR reflects the API structure of My Account API (which can be found [here](https://github.com/atko-cic/my-account/blob/73fe761f3695346d6ae6f4c1dcd8f97ff446c881/oas/schema/components/request_bodies/connected_accounts/CreateConnectedAccountsRequestContent.json) and has a top level scopes parameter.

> A note that the new scopes field is now an array of strings, which is a meaningful API change.

This again has been debated but ultimately this has been decided because it provides a better DX for passing scopes to the upstream rather than having to deal with the string manipulation as a developer. Whereas the `scope` parameter as a string is constrained by legacy and OAuth2 in general requiring `scope` as a string - we do not have the same constraints here. `scopes` here is actually use in quite a different way than elsewhere.

> Clarification on usage: Should users still pass scope inside authorizationParams? The type definition still allows it, so it would be good to explain how that scope differs from the new top-level scopes, and which one should be preferred going forward.

Yes and no. They should not be used in a connected account connect request. The authorizationParams.scope field is not used by MyAccount (they are still present on the OpenAPI spec definition to avoid immediately breaking everyone but they are completely ignored). `scope` used on authorizationParams elsewhere (e.g. Login) or on the default ones defined at configuration is still valid.

PR #1477, opened 2025-12-04
feat: add organization parameter support to token exchange

yogeshchoudhary147:
## Description
Adds support for the `organization` parameter in the token exchange flow (`exchangeToken` method) to enable organization-specific authentication.

## Changes
- Added `organization` as an optional parameter to `CustomTokenExchangeOptions` interface
- Updated `exchangeToken` method to conditionally pass organization to the `/oauth/token` endpoint
- Updated JSDoc documentation with organization parameter details and usage example
- Added test coverage for organization parameter (both provided and not provided scenarios)

## API Changes
The `/oauth/token` endpoint now accepts an optional `organization` parameter for the token exchange grant type. When provided, the organization ID will be present in the access token payload.

## Testing
- ✅ Added test: passes organization parameter to _requestToken when provided
- ✅ Added test: does not pass organization parameter when not provided
- ✅ All existing tests pass

## Example Usage
```typescript
const tokenResponse = await auth0Client.exchangeToken({
  subject_token: 'external_token',
  subject_token_type: 'urn:acme:legacy-system-token',
  scope: 'openid profile',
  organization: 'org_12345'
});
// Organization ID will be in access token payload
```


src/Auth0Client.ts
```diff
@@ -1454,7 +1458,8 @@ export class Auth0Client {
         options.scope,
         options.audience || this.options.authorizationParams.audience
       ),
-      audience: options.audience || this.options.authorizationParams.audience
+      audience: options.audience || this.options.authorizationParams.audience,
+      ...(options.organization && { organization: options.organization })
```
yogeshchoudhary147, line 1462: Should I update `exchangeToken` to use the fallback pattern for consistency, or is there a reason to keep it as-is (only using the explicitly passed organization)?

**Current implementation:**

`...(options.organization && { organization: options.organization })`

**Proposed change:**

`organization: options.organization || this.options.authorizationParams.organization`

gyaneshgouraw: The suggested approach also works fine, i don't see any issues with `organization` following the existing pattern.

PR #1500, opened 2026-01-09
fix: add closePopup option for Chrome extension compatibility

### LINKED ISSUE #1318: Login popup closes before flow is complete, making login impossible in Chrome extensions
### Checklist

- [X] The issue can be reproduced in the [auth0-spa-js sample app](https://github.com/auth0-samples/auth0-javascript-samples/tree/master/01-Login) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/auth0-spa-js#readme), [Examples](https://github.com/auth0/auth0-spa-js/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-spa-js/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://auth0.com/docs/libraries/auth0-single-page-app-sdk) and [API documentation](https://auth0.github.io/auth0-spa-js/), and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-spa-js/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

`loginWithPopup` closes the popup is closed before the token is requested. When used in a Chrome extension, this causes the extension to be closed as well because the browser window regains focus. Closing the extension interrupts the script, so it is shut down without ever requesting the token or saving it to the cache.

Because the token wasn't requested by the extension before being forcibly closed, the cookie is not set and the user is not authenticated when they re-open the extension. This makes it impossible for the user to ever sign in.

I've done some testing and simply delaying the call to `config.popup.close()` until the very end of `loginWithPopup` fixes things. The extension still closes upon successful sign in, but once re-opened the user will be authenticated.

I will be opening a PR to fix this shortly. I think the best approach it to allow suppressing the `popup.close()` call entirely, so devs can manually close it at whatever point in their login flow is an okay point for the extension to be forcibly closed.

### Reproduction

1. Open Chrome extension
2. Click button that triggers `loginWithPopup`
3. Complete login in the popup
4. Popup & extension both close at the same time
5. Re-open the extension
6. Call `checkSession`
7. See that user is not authenticated

If you open devtools for the extension prior to calling `loginWithPopup`, the extension will not close and the login flow will work as expected.

This can be reproduced consistently for me on Chrome 130 on Fedora.

### Additional context

This issue has been reported before, without a real solution or understanding of why it was happening:
- https://community.auth0.com/t/ultimate-guide-to-auth0-in-a-chrome-extension-popup/61362/2
- https://github.com/auth0/auth0-spa-js/issues/567

Regarding Chrome closing the extension when the window regains focus after the popup closes, see [this docs page](https://developer.chrome.com/docs/extensions/develop/ui/add-popup):

> There is no way to keep the [extension] popup open after the user has clicked away.

This is arguably a very poor UX choice in the context of popups, but not something we can change so we just have to work around it.

We also can't just use `loginWithRedirect`, as redirects don't work in an extension.

### auth0-spa-js version

2.1.3

### Which framework are you using (React, Angular, Vue...)?

Next.js

### Framework version

14.2.3

### Which browsers have you tested in?

Chrome

yogeshchoudhary147:
Fixes #1318

## Problem
When using `loginWithPopup()` in Chrome extensions, the popup closes immediately after receiving the authorization response. This terminates the extension's service worker before token exchange completes, leaving the user unauthenticated.

## Solution
Added a `closePopup` option to `PopupConfigOptions`:
- `true` (default): SDK closes the popup automatically after receiving the authorization response
- `false`: SDK does not close the popup. The caller is responsible for closing it, including on errors.

This gives users full control over the popup lifecycle, which is especially useful for Chrome extensions where closing too early terminates the service worker.

## Changes
- **utils.ts**: Modified popup closing logic to respect `closePopup` option
- **global.ts**: Added JSDoc documentation for `closePopup` option with try/finally usage example
- **loginWithPopup.test.ts**: Added comprehensive test coverage for the `closePopup` option

## Usage
```typescript
const popup = window.open('', '_blank');
try {
  await auth0.loginWithPopup({}, {
    popup: popup,
    closePopup: false  // SDK won't close the popup
  });
  // Tokens are cached, authentication successful
} catch (error) {
  // Handle authentication errors
  console.error('Login failed:', error);
} finally {
  // Always close popup, even on errors
  popup.close();
}
```

This also provides flexibility for other use cases like showing post-login messages or executing additional logic before closing the popup.

### Unit Tests
Added 4 test cases covering:
- Default behavior (popup closes immediately)
- Explicit `closePopup: true`
- `closePopup: false` (SDK does not close popup on success)
- `closePopup: false` (SDK does not close popup on token exchange failure)

All tests pass ✅

### Manual Testing
Manually tested the fix in Chrome extension scenarios:

**Before fix:**
- ❌ Popup closes immediately after authorization
- ❌ Extension terminates before token exchange
- ❌ User remains unauthenticated (no tokens cached)

**After fix with `closePopup: false`:**
- ✅ SDK does not close the popup
- ✅ User closes popup in finally block after `loginWithPopup` completes
- ✅ Tokens are successfully cached before popup closes
- ✅ User is authenticated
- ✅ On reopening extension, user is logged in (tokens persisted in localStorage)
- ✅ Popup properly closed even when errors occur

**Other scenarios:**
- ✅ Default behavior: Popup closes immediately after authorization (no regression)

## Backward Compatibility
No breaking changes. Default behavior unchanged - popup still closes immediately after authorization.

## Credits
This PR builds upon the initial investigation and work by @ashwsn.

Closes #1318

### DIFF
--- a/__tests__/Auth0Client/loginWithPopup.test.ts
+++ b/__tests__/Auth0Client/loginWithPopup.test.ts
@@ -811,36 +811,6 @@ describe('Auth0Client', () => {
       expect(popup.close).toHaveBeenCalled();
     });
 
-    it('should close popup on state mismatch error regardless of closePopup setting', async () => {
-      const auth0 = setup();
-      const popup = {
-        location: { href: '' },
-        close: jest.fn()
-      };
-
-      let error;
-      try {
-        await loginWithPopup(
-          auth0,
-          {},
-          { popup, closePopup: false },
-          {
-            authorize: {
-              response: {
-                state: 'other-state'
-              }
-            }
-          }
-        );
-      } catch (e) {
-        error = e;
-      }
-
-      expect(error).toBeDefined();
-      expect(error.message).toBe('Invalid state');
-      expect(popup.close).toHaveBeenCalled();
-    });
-
     it('should not close popup when closePopup is false', async () => {
       const auth0 = setup();
       const popup = {
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -429,10 +429,6 @@ export class Auth0Client {
     });
 
     if (params.state !== codeResult.state) {
-      // Always close popup on state mismatch error, regardless of closePopup setting
-      if (config.popup) {
-        config.popup.close();
-      }
       throw new GenericError('state_mismatch', 'Invalid state');
     }
 
--- a/src/global.ts
+++ b/src/global.ts
@@ -409,11 +409,21 @@ export interface PopupConfigOptions {
    * Controls whether the SDK automatically closes the popup window.
    *
    * - `true` (default): SDK closes the popup automatically after receiving the authorization response
-   * - `false`: SDK does not close the popup. The caller is responsible for closing it.
+   * - `false`: SDK does not close the popup. The caller is responsible for closing it, including on errors.
    *
-   * Setting this to `false` is useful when you need control over when the popup closes,
+   * Setting this to `false` is useful when you need full control over the popup lifecycle,
    * such as in Chrome extensions where closing the popup too early can terminate the
    * extension's service worker before authentication completes.
+   *
+   * When `closePopup: false`, you should close the popup in a try/finally block:
+   * ```
+   * const popup = window.open('', '_blank');
+   * try {
+   *   await auth0.loginWithPopup({}, { popup, closePopup: false });
+   * } finally {
+   *   popup.close();
+   * }
+   * ```
    *
    * @default true
    */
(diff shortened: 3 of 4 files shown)


frederikprijck (changes requested):
A boolean called closePopup  to control _when_  and not _if_ a popup should be closed  feels weird. I expect when setting it to false, it wouldnt close, so this code feels very unexpected:

```
if (config.closePopup === false && config.popup) {
  config.popup.close();
}
```

PR #1505, opened 2026-01-13
fix: add two-tier locking to prevent iframe race conditions causing "Invalid state" errors

yogeshchoudhary147:
## Problem
Customers experiencing "Invalid state" errors after upgrading to v2.8.0 when calling `getTokenSilently()` for multiple audiences in parallel with iframe-based authentication.

**Root Cause**: PR #1408 changed locking from global to per-audience to enable MRRT parallelization. This allows concurrent iframes to overwrite each other's state parameter in Auth0's server-side session.

**Reproduces**: v2.8.0+ ✅ | v2.7.0 ❌

## Demo Video
Here's a video demonstrating the bug and the fix:

https://github.com/user-attachments/assets/24d80ea9-4ae0-4d8c-a751-8fb970239ba5

**What the video shows:**
- Without the fix (v2.8.0): "Invalid state" errors when making parallel iframe calls
- With the fix: Both parallel iframe calls succeed with two-tier locking

## Solution
Implement two-tier locking:
- **Outer lock (per-audience)**: Prevents duplicate calls for same audience → preserves MRRT performance
- **Inner lock (global iframe)**: Serializes iframe authorization flows → prevents session state conflicts

Refresh token flows bypass the inner lock and remain fully parallel.

## Changes
- Added `buildIframeLockKey()` to create global iframe lock key
- Modified `_getTokenFromIFrame()` to acquire/release lock with try-finally guarantee
- Zero breaking changes

## Impact
- ✅ Fixes "Invalid state" regression from v2.8.0
- ✅ Maintains MRRT performance improvements
- ✅ Affects: `useRefreshTokens: false` or `useRefreshTokensFallback: true` scenarios

## Testing
Tested with multiple parallel `getTokenSilently()` calls for different audiences - no "Invalid state" errors observed.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,2 @@
 docs
 static/auth0-spa-js.development_old.js
-src/Auth0Client.ts


src/Auth0Client.ts
```diff
@@ -1023,32 +1024,44 @@ export class Auth0Client {
...
-      prompt: 'none'
-    };
+    const iframeLockKey = buildIframeLockKey(this.options.clientId);
 
-    const orgHint = this.cookieStorage.get<string>(this.orgHintCookieName);
+    // Acquire global iframe lock to serialize iframe authorization flows
+    const lockAcquired = await retryPromise(
+      () => lock.acquireLock(iframeLockKey, 5000),
+      10
+    );
```
frederikprijck: Can we be a bit more descriptive on _why_ we are doing that?

```suggestion
    // Acquire global iframe lock to serialize iframe authorization flows
    // This is necessary because the SDK does not support multiple simultaneous transactions.
    // Since https://github.com/auth0/auth0-spa-js/pull/1408, when calling `getTokenSilently()`, the global locking will lock per `audience` instead of locking only per `client_id`.
    // This means that calls for different audience would happen in parallel, which does not work when using silent authentication (prompt=none) from within the SDK, as that relies on the same transaction context as a top-level `loginWithRedirect`.
    // To resolve that, we add a second-level locking that locks only the iframe calls in the same way as was done before changing this in https://github.com/auth0/auth0-spa-js/pull/1408.
    const lockAcquired = await retryPromise(
      () => lock.acquireLock(iframeLockKey, 5000),
      10
    );
```

frederikprijck: Above is just a proposal, feel free to change.

src/Auth0Client.utils.ts
```diff
@@ -21,6 +21,16 @@ export const buildGetTokenSilentlyLockKey = (
...
 
+/**
+ * @ignore
+ * Builds a global lock key for iframe-based authentication flows.
+ * This ensures only one iframe authorization request runs at a time per client,
+ * preventing "Invalid state" errors from concurrent iframe requests overwriting
+ * each other's state in the Auth0 session.
+ */
+export const buildIframeLockKey = (clientId: string) =>
+  `auth0.lock.iframe.${clientId}`;
```
frederikprijck: To be consistent, should we add a `GET_TOKEN_FROM_IFRAME_LOCK_KEY`, set to `auth0.lock.getTokenFromIFrame`?

PR #1540, opened 2026-02-20
feat: support MFA step-up authentication via popup in iframe-based silent auth flow

gyaneshgouraw-okta:
### Summary

Extend `interactiveErrorHandler: 'popup'` to handle MFA step-up errors from the iframe path. Previously, step-up only worked with refresh tokens. The iframe path returned `login_required` (not `mfa_required`), which caused a premature `logout()` and was not detected as an interactive error. Now both token acquisition paths trigger the popup automatically.

Extends step-up functionality added in https://github.com/auth0/auth0-spa-js/pull/1531

### Changes

- Detect iframe MFA step-up errors by matching `error='login_required'` + `error_description='Multifactor authentication required'`
- Widen interactive error detection to match both `MfaRequiredError` (refresh tokens) and the iframe `GenericError` variant
- Skip `logout()` when the `login_required` error is an MFA step-up and the handler is configured, preserving the session for the popup
- Add two tests: popup opens on iframe MFA error, and `logout()` is not called
- Remove internal jargon ("iframe flow", "refresh token flow") from user-facing step-up docs

### Example

```js
// Works with or without useRefreshTokens — no change needed
  domain: '<AUTH0_DOMAIN>',
  interactiveErrorHandler: 'popup'
});

// If MFA is required, popup opens automatically
const token = await auth0.getTokenSilently({
  authorizationParams: {
    audience: 'https://api.example.com',
    scope: 'read:sensitive-data'
  }
});
```

### Testing

- Added two new tests to cover the recent changes  
- All tests are passing  
- `npm run build` completes successfully  
- Performed manual testing using `static/step-up.html`: reproduced the iframe-based silent token flow with MFA step-up and confirmed that the popup appears and successfully returns a token

### DIFF
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -1233,7 +1233,7 @@ export class Auth0Client {
         const shouldSkipLogoutForMfaStepUp =
           e instanceof GenericError &&
           this._isIframeMfaError(e) &&
-          this.options.interactiveErrorHandler;
+          this.options.interactiveErrorHandler === 'popup';
 
         if (!shouldSkipLogoutForMfaStepUp) {
           this.logout({


src/Auth0Client.ts
```diff
@@ -1207,9 +1227,19 @@ export class Auth0Client {
...
-        this.logout({
-          openUrl: false
-        });
+        // When the login_required error is actually an MFA step-up requirement
+        // and the interactive error handler is configured, skip logout so the
+        // session is preserved for the popup flow.
+        const shouldSkipLogoutForMfaStepUp =
+          e instanceof GenericError &&
+          this._isIframeMfaError(e) &&
+          this.options.interactiveErrorHandler;
```
yogeshchoudhary147: Shouldn't this be `this.options.interactiveErrorHandler === 'popup'`?

gyaneshgouraw: Using just `this.options.interactiveErrorHandler` (truthy check) works the same today since 'popup' is the only supported value, but `=== 'popup'` check is more explicit.  Updated the code to use the explicit check.

gyaneshgouraw: Updated the check in this commit - https://github.com/auth0/auth0-spa-js/pull/1540/changes/b318cdd6b5bd702ea9fc9694e8858fdd64d36f14

PR #1541, opened 2026-02-22
feat: add refresh token revocation

yogeshchoudhary147:
## Summary

Adds `revokeRefreshToken()` to `Auth0Client` to explicitly revoke refresh tokens via `/oauth/revoke` (RFC 7009).

- Revokes all distinct refresh tokens for the given audience in one call
- Preserves the access token in cache — only `refresh_token` is cleared
- MRRT-aware: revoking a shared token cleans up all associated cache entries
- Worker path: tokens never leave the worker thread
- Scope excluded from the API — RTs are not scope-bound
- Non-worker path now revokes and strips tokens incrementally, so partial failures keep cache state consistent for already-revoked tokens

## Usage

```js
await auth0.revokeRefreshToken();
await auth0.revokeRefreshToken({ audience: 'https://api.example.com' });
```

> For full session termination use `logout()` instead.

## Test plan

- [x] `Auth0Client.revokeRefreshToken()`: success, no-op, cache cleanup, AT preservation, error handling, audience override, worker path, multi-RT, fail-fast
- [x] `Auth0Client.revokeRefreshToken()`: partial-failure regression (first RT revoked+stripped, second RT fails, cache remains consistent)
- [x] `revokeMessageHandler`: multi-RT loop, MRRT cleanup, timeout, form-data, fail-fast
- [x] `revokeToken` (non-worker + worker paths): JSON/form-data, error responses, timeout, malformed body
- [x] `getRefreshTokensByAudience`: single RT, MRRT dedup, multiple distinct RTs, audience isolation
- [x] `stripRefreshToken`: MRRT cleanup, partial match
- [x] Manual testing against a real Auth0 tenant

### DIFF
--- a/__tests__/api.test.ts
+++ b/__tests__/api.test.ts
@@ -657,4 +657,72 @@ describe('revokeToken', () => {
       })
     ).rejects.toThrow('HTTP error 400');
   });
+
+  it('delegates to the worker via sendMessage when a worker is provided', async () => {
+    const workerUtils = require('../src/worker/worker.utils');
+    jest.spyOn(workerUtils, 'sendMessage').mockResolvedValue(undefined);
+
+    const worker = {} as Worker;
+
+    await revokeToken(
+      {
+        baseUrl: `https://${TEST_DOMAIN}`,
+        client_id: TEST_CLIENT_ID,
+        refreshTokens: [],
+        auth0Client: DEFAULT_AUTH0_CLIENT,
+        audience: '__test_audience__'
+      },
+      worker
+    );
+
+    expect(workerUtils.sendMessage).toHaveBeenCalledWith(
+      {
+        type: 'revoke',
+        timeout: 10000,
+        fetchUrl: `https://${TEST_DOMAIN}/oauth/revoke`,
+        fetchOptions: {
+          method: 'POST',
+          body: JSON.stringify({ client_id: TEST_CLIENT_ID, token_type_hint: 'refresh_token' }),
+          headers: {
+            'Content-Type': 'application/json',
+            'Auth0-Client': btoa(JSON.stringify(DEFAULT_AUTH0_CLIENT))
+          }
+        },
+        useFormData: undefined,
+        auth: { audience: '__test_audience__' }
+      },
+      worker
+    );
+  });
+
+  it('sends form-encoded body to worker when useFormData is true', async () => {
+    const workerUtils = require('../src/worker/worker.utils');
+    jest.spyOn(workerUtils, 'sendMessage').mockResolvedValue(undefined);
+
+    const worker = {} as Worker;
+
+    await revokeToken(
+      {
+        baseUrl: `https://${TEST_DOMAIN}`,
+        client_id: TEST_CLIENT_ID,
+        refreshTokens: [],
+        auth0Client: DEFAULT_AUTH0_CLIENT,
+        useFormData: true,
+        audience: '__test_audience__'
+      },
+      worker
+    );
+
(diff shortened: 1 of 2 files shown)


src/http.ts
```diff
@@ -227,3 +228,55 @@ export async function getJSON<T>(
...
+    fetch(fetchUrl, fetchOptions),
+    new Promise<never>((_, reject) => {
+      timeoutId = setTimeout(() => {
+        controller.abort();
+        reject(new Error("Timeout when executing 'fetch'"));
+      }, timeout);
+    })
+  ]).finally(() => {
+    clearTimeout(timeoutId);
+  });
```
frederikprijck: nit: should we consider making this more readable by using something like below?

```
const response = await fetchWithTimeout(fetchUrl, fetchOptions, timeout)
```

And then implement it like:

```
function fetchWithTimeout(input, init, timeout) {
  let timeoutId: NodeJS.Timeout;
  const controller = createAbortController();

  init.signal = controller.signal;

  return await Promise.race([
    fetch(input, {
      signal: controller.signal,
      ...init
    }),
    new Promise<never>((_, reject) => {
      timeoutId = setTimeout(() => {
        controller.abort();
        reject(new Error("Timeout when executing 'fetch'"));
      }, timeout);
    })
  ]).finally(() => {
    clearTimeout(timeoutId);
  });
}
```

yogeshchoudhary147: Done, extracted it into a `fetchWithTimeout` helper.

src/worker/token.worker.ts
```diff
@@ -181,8 +203,84 @@ const messageHandler = async ({
...
+      }
+
+      let timeoutId: ReturnType<typeof setTimeout>;
+      let response: void | Response;
+
+      try {
+        response = await Promise.race([
+          new Promise<void>(resolve => { timeoutId = setTimeout(resolve, timeout); }),
+          fetch(fetchUrl, { ...fetchOptions, body, signal })
+        ]).finally(() => clearTimeout(timeoutId));
```
tusharpandey13, line 245: Not an issue but a reccomendation:
`finally` here cleanly does a `clearTimeout` which destroys the timer.
In the same file, (https://github.com/auth0/auth0-spa-js/blob/addeda65f2216ee0c6d9c02e5f399a44947b09d6/src/worker/token.worker.ts#L144) for the existing `messageHandler`, there is no `finally` so that timer is cleared only by GC.

Again, not an issue with current code but an existing memory leakage can be easily fixed here 🙂

src/api.ts
```diff
@@ -59,3 +82,89 @@ export async function oauthToken(
...
+
+    if (!response.ok) {
+      let errorDescription: string | undefined;
+      try {
+        const { error_description } = JSON.parse(await response.text());
+        errorDescription = error_description;
+      } catch {
+        // body absent or not valid JSON
+      }
+      throw new Error(errorDescription || `HTTP error ${response.status}`);
```
tusharpandey13: Maybe we can throw a Auth0 typed `GenericError` here, so that error info is available after sendMessage:
```
throw new GenericError(                                                                                                                        
      error || 'revoke_error',                                                                                                                     
      errorDescription || `HTTP error ${response.status}`                                                                                          
    );
```

yogeshchoudhary147: Fixed, thanks!

src/worker/token.worker.ts
yogeshchoudhary147: Good catch! Will fix this in a separate PR to keep this one focused.

PR #1547, opened 2026-02-23
Feat/native to web sso support

nelsonmaia:
Changes

  This PR adds automatic Native to Web SSO support to the SDK, enabling seamless single sign-on when users transition from a native mobile application to a web application.

  New configuration option:
  - enableSessionTransfer?: boolean (default: true) - Added to Auth0ClientOptions in src/global.ts

  New typed property:
  - session_transfer_token?: string - Added to AuthorizationParams in src/global.ts

  Core logic:
  - Added _extractSessionTransferToken() private method to Auth0Client that extracts session_transfer_token from URL query parameters
  - Modified _prepareAuthorizeUrl() to automatically include the session transfer token in /authorize requests when present in URL

  Usage:
  ```
// Automatic (default behavior):
  // When web app is opened with URL: https://myapp.com?session_transfer_token=xyz
  await auth0.loginWithRedirect();
  // → /authorize request automatically includes session_transfer_token=xyz

  // To disable:
    domain: '...',
    clientId: '...',
    enableSessionTransfer: false
  });

  // Manual override (takes precedence over URL):
  await auth0.loginWithRedirect({
    authorizationParams: {
      session_transfer_token: 'manual-token'
    }
  });

```
  References

  - https://auth0.com/docs/authenticate/single-sign-on/native-to-web
  - https://auth0.com/docs/authenticate/single-sign-on/native-to-web/configure-implement-native-to-web

  Testing

  - 8 unit tests added in __tests__/Auth0Client/sessionTransfer.test.ts covering:
    - Auto-detection when enabled (default)
    - Auto-detection when explicitly enabled
    - No detection when disabled (enableSessionTransfer: false)
    - Manual token takes precedence over URL token
    - No token when not present in URL
    - Mixed query params handling
    - URL-encoded token handling
    - Popup flow compatibility

  To test manually:
  1. Configure a native app with can_create_session_transfer_token: true
  2. Configure a web app with allowed_authentication_methods: ["query"]
  3. Open web app with ?session_transfer_token=<token> in URL

### Checklist

- [ ] All code quality tools/guidelines have been run/followed

### DIFF
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -386,13 +386,8 @@ export class Auth0Client {
    * @returns The session transfer token if present, undefined otherwise
    */
   private _extractSessionTransferToken(paramName: string): string | undefined {
-    if (typeof window === 'undefined') return;
-    try {
-      const params = new URLSearchParams(window.location.search);
-      return params.get(paramName) || undefined;
-    } catch {
-      return;
-    }
+    const params = new URLSearchParams(window.location.search);
+    return params.get(paramName) || undefined;
   }
 
   /**
@@ -403,7 +398,6 @@ export class Auth0Client {
    * @param paramName The query parameter name to remove from the URL
    */
   private _clearSessionTransferTokenFromUrl(paramName: string): void {
-    if (typeof window === 'undefined') return;
     try {
       const url = new URL(window.location.href);
       if (url.searchParams.has(paramName)) {

### CODE REVIEW

File: EXAMPLES.md [hunk 1]
```diff
@@ -289,6 +290,75 @@ if (organization && invitation) {
...
+// https://yourapp.com?session_transfer_token=xyz123
+//
+// The SDK automatically includes the token in loginWithRedirect:
+await auth0.loginWithRedirect();
+// The /authorize request will include session_transfer_token=xyz123
+```
+
+### Disabling Automatic Detection
+
+If you need to disable automatic session transfer token detection, set `enableSessionTransfer` to `false`:
```
frederikprijck: `enableSessionTransfer` only toggles the automatic parsing of the URL. It still works fine with setting it explicitly when calling `loginWithRedirect`, regardless of the value of `enableSessionTransfer`.

To me that reads a bit confusing. I think this flag is specific to the URL parsing only, and we should make that clear.

nelsonmaia: update it, let me know what do you think

frederikprijck: I think my main concern was about the name of the property, not the docs specifically. But that was probably not clear from my comment, apologies! However, I do like the improvements to the docs 👍

src/Auth0Client.ts
```diff
@@ -378,6 +377,22 @@ export class Auth0Client {
...
+   * Extracts the session_transfer_token from the current URL query parameters
+   * for Native to Web SSO flows.
+   *
+   * @returns The session transfer token if present, undefined otherwise
+   */
+  private _extractSessionTransferToken(): string | undefined {
+    if (typeof window === 'undefined') return;
+    try {
+      const params = new URLSearchParams(window.location.search);
+      return params.get('session_transfer_token') || undefined;
```
yogeshchoudhary147: Unlike `loginWithRedirect` (where the browser navigates away and the URL changes naturally), `loginWithPopup` leaves the main page URL unchanged. Could the stale `session_transfer_token` persist and get re-sent on the next silent refresh?

src/global.ts
```diff
@@ -324,6 +333,22 @@ export interface Auth0ClientOptions {
...
   authorizationParams?: ClientAuthorizationParams;
+
+  /**
+   * If `true`, the SDK automatically extracts `session_transfer_token` from URL query
+   * parameters and includes it in authorization requests for Native to Web SSO.
+   *
+   * This enables seamless single sign-on when users transition from a native mobile
+   * application to a web application.
+   *
+   * The default setting is `true`.
```
yogeshchoudhary147: Any existing app upgrading that happens to use `session_transfer_token` as its own URL param will silently start forwarding it to Auth0. 

Chances are low but should this default to `false`? Generally new behavior in a minor version should be opt-in.

__tests__/Auth0Client/sessionTransfer.test.ts
```diff
@@ -0,0 +1,219 @@
...
+
+import {
+  TEST_CLIENT_ID,
+  TEST_CODE_CHALLENGE,
+  TEST_DOMAIN,
+  TEST_NONCE,
+  TEST_REDIRECT_URI,
+  TEST_SCOPES,
+  TEST_STATE
+} from '../constants';
```
yogeshchoudhary147, line 12: `assertUrlEquals`, `TEST_CLIENT_ID`, `TEST_NONCE`, `TEST_REDIRECT_URI`, `TEST_SCOPES` are imported but never used.

__tests__/Auth0Client/sessionTransfer.test.ts
```diff
@@ -0,0 +1,219 @@
...
+        });
+
+        // We can't easily verify popup URL in this test setup,
+        // but the same _prepareAuthorizeUrl is used
+        // This test ensures no errors occur with session transfer in popup flow
+        try {
+          await auth0.loginWithPopup();
+        } catch {
+          // Expected to fail due to mock setup, but no session transfer errors
+        }
```
yogeshchoudhary147, line 440: There is no expectation here. The popup URL is accessible via the mock popup object returned by `mockWindow.open`.  You can assert on it after the catch block.

EXAMPLES.md
nelsonmaia: @frederikprijck 
Based on what we usually use in our SDKs, which one would you prefer or if you have any better suggestion too:

`autoParseSessionTransferToken`
`parseSessionTransferTokenFromUrl`
`useSessionTransferTokenFromUrl`
`autoExtractSessionTransferToken`

src/global.ts
nelsonmaia: I've discussed this with @frederikprijck , and since this is a very unusual name of parameters and it would only have any effect in Auth0 , if N2W is configured there, I think that leaving it true makes more sense and improves the DX , but I'm open to change to `false` and wait the next major version. 

I'm not sure how would be the process of changing in the next major, would I have to be aware of that?

frederikprijck: Typically, new behavior gets added as opt-in first, then opt-out in a new major. And if needed, in another major it can be conmpletely removed.

src/Auth0Client.ts
nelsonmaia: You're right! The stale token in the URL could get re-sent during silent auth, which shouldn't happen.

  I'll move the session_transfer_token extraction out of _prepareAuthorizeUrl() and into loginWithRedirect() and loginWithPopup()
  only. This way silent auth won't pick it up.

  Does that solution work for you?

src/global.ts
frederikprijck: I realy think it should not be true to be honest. Even more so, I am not convinced on the automagic behavior of sending query parameters to auth0.

nelsonmaia: We've agreed on opt-in approach and the configuration will be 

```
{
  sessionTransferToken: {
    mode: 'auto',  // optional default to manual
    queryParamName: 'my_custom_name'  // optional default to session_transfer_token
  }
}
```

src/Auth0Client.ts
yogeshchoudhary147: For `loginWithPopup`, the URL persists so the token would keep getting re-sent on subsequent calls. Would it make sense to clear it from the URL after extraction?

src/global.ts
yogeshchoudhary147: After team discussion, can we replace `enableSessionTransfer: boolean` with a single `sessionTransferTokenQueryParamName: string`? Setting it enables auto-extraction, leaving it unset is the no-op default.

src/Auth0Client.ts
nelsonmaia: it makes, I just implemented in the latest commit

src/global.ts
nelsonmaia: @yogeshchoudhary147 just commit, can you review when you have a chance

src/global.ts
```diff
@@ -113,6 +113,15 @@ export interface AuthorizationParams {
    */
   redirect_uri?: string;
 
+  /**
+   * Session transfer token from a native application for Native to Web SSO.
+   * When `enableSessionTransfer` is true (default), this is automatically
```
yogeshchoudhary147: A comment still references `enableSessionTransfer` which no longer exists. Should be updated to reference `sessionTransferTokenQueryParamName`.

package-lock.json
```diff
@@ -182,6 +182,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
```
yogeshchoudhary147: These appear unrelated to the feature.

src/Auth0Client.ts
```diff
@@ -462,8 +499,30 @@ export class Auth0Client {
...
+      if (sessionTransferToken) {
+        authorizationParams = {
+          ...authorizationParams,
+          session_transfer_token: sessionTransferToken
+        };
+        // Clear the token from URL to prevent re-use on subsequent calls
+        this._clearSessionTransferTokenFromUrl(
+          this.options.sessionTransferTokenQueryParamName
+        );
+      }
```
yogeshchoudhary147: This block is duplicated verbatim in both `loginWithPopup` and `loginWithRedirect`. Could you extract it into a private helper, e.g.:

```ts
private _applySessionTransferToken(
  authorizationParams: AuthorizationParams
): AuthorizationParams {
  const paramName = this.options.sessionTransferTokenQueryParamName;
  if (!paramName || authorizationParams.session_transfer_token) {
    return authorizationParams;
  }
  const token = this._extractSessionTransferToken(paramName);
  if (!token) return authorizationParams;
  this._clearSessionTransferTokenFromUrl(paramName);
  return { ...authorizationParams, session_transfer_token: token };
}
```

Then both call sites become a single line:
```ts
const authorizationParams = this._applySessionTransferToken(options.authorizationParams || {});
```

frederikprijck (approved):

PR #1556, opened 2026-03-11
build(deps): pin dependency versions in package.json

gyaneshgouraw-okta:
## Summary

After removing `package-lock.json` from source control https://github.com/auth0/auth0-spa-js/pull/1551, semver ranges  allowed different CI runs on the same commit to resolve different dependency versions, producing non-deterministic bundles. Since all runtime deps are bundled into the final output via Rollup, pinning ensures the shipped artifact is identical across builds.

## Changes

- Removed semver range prefixes  from all 4 runtime dependencies: `@auth0/auth0-auth-js`, `browser-tabs-lock`, `dpop`, `es-cookie`
- Dependency upgrades would continue to flow through Dependabot PRs  no change to the update workflow

## Testing

- `npm run build` produces the same bundle output as before
-  `npm run test` passes all test

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@auth0/auth0-auth-js": "1.5.0",
-    "browser-tabs-lock": "1.2.15",
+    "browser-tabs-lock": "1.3.0",
     "dpop": "2.1.1",
     "es-cookie": "1.3.2"
   },


package.json
```diff
@@ -23,10 +23,10 @@
...
   },
   "dependencies": {
-    "@auth0/auth0-auth-js": "^1.5.0",
-    "browser-tabs-lock": "^1.2.15",
-    "dpop": "^2.1.1",
-    "es-cookie": "~1.3.2"
+    "@auth0/auth0-auth-js": "1.5.0",
+    "browser-tabs-lock": "1.2.15",
+    "dpop": "2.1.1",
+    "es-cookie": "1.3.2"
```
frederikprijck, line 29: Are we sure this is the correct pinning? I can see our lock file used to contain:

- `es-cookie` locked to 1.3.2: https://github.com/auth0/auth0-spa-js/blob/fbfcb2d27450888c96b750fc6938cf00d37ca5d6/package-lock.json#L6866
- `dpop` locked to 2.1.1: https://github.com/auth0/auth0-spa-js/blob/fbfcb2d27450888c96b750fc6938cf00d37ca5d6/package-lock.json#L6601

But

- `browser-tabs-lock` locked to 1.3.0: https://github.com/auth0/auth0-spa-js/blob/fbfcb2d27450888c96b750fc6938cf00d37ca5d6/package-lock.json#L4462
- `@auth0/auth0-auth-js` locked to 1.4.0: https://github.com/auth0/auth0-spa-js/blob/fbfcb2d27450888c96b750fc6938cf00d37ca5d6/package-lock.json#L97

gyaneshgouraw: Thanks for highlighting the inconsistency in pinning version for  `browser-tabs-lock`.
Reviewed it and updated to correct version which was used earlier to `1.3.0` in this commit - https://github.com/auth0/auth0-spa-js/pull/1556/changes/057b901cad52e723c18b5935be6d3226028155f3

Regarding `@auth0/auth0-auth-js`
It's correctly locked  `1.5.0` which was used earlier also.

yogeshchoudhary147 (approved):

PR #1588, opened 2026-04-11
docs: add revokeRefreshToken examples

yogeshchoudhary147:
## Summary

- Documents `revokeRefreshToken()` in EXAMPLES.md under the Refresh Tokens section
- Covers basic usage, cache semantics, difference from `logout()`, error handling, multi-audience pattern, multiple RTs per audience (scope combinations), and MRRT behavior

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -155,23 +155,19 @@ This method only has an effect when `useRefreshTokens` is `true`. If refresh tok
 await auth0.revokeRefreshToken();
 ```
 
-**How it affects the cache:** The access token is preserved in the cache — only the refresh token entry is cleared. The next call to `getTokenSilently()` will require a new login if the access token has also expired.
+**How it affects the cache:** The access token is preserved in the cache — only the refresh token entry is cleared. Once the access token expires, `getTokenSilently()` will attempt silent auth (via iframe, if `useRefreshTokensFallback` is enabled and the Auth0 session is still active) before requiring a new interactive login.
 
 **Difference from `logout()`:** `revokeRefreshToken()` invalidates the refresh token on the Auth0 server and removes it from the local cache, but it does **not** clear the user's Auth0 session or the rest of the local cache. If you want to fully terminate the session, use `logout()` instead.
 
 #### Error Handling
 
-`revokeRefreshToken()` throws a `GenericError` if the `/oauth/revoke` endpoint returns an error (for example, if the token has already been revoked or is invalid). Wrap the call in a try/catch:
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -144,6 +144,64 @@ await createAuth0Client({
...
+The `revokeRefreshToken()` method explicitly revokes a refresh token via the `/oauth/revoke` endpoint ([RFC 7009](https://www.rfc-editor.org/rfc/rfc7009)). This invalidates the refresh token so it can no longer be used to obtain new access tokens.
+
+This method only has an effect when `useRefreshTokens` is `true`. If refresh tokens are disabled it returns immediately without doing anything.
+
+```js
+// Revoke the refresh token for the default audience
+await auth0.revokeRefreshToken();
+```
+
+**How it affects the cache:** The access token is preserved in the cache — only the refresh token entry is cleared. The next call to `getTokenSilently()` will require a new login if the access token has also expired.
```
nandan-bhat: Should we make this sentence a bit more careful ?

After the access token expires, `getTokenSilently()` does not always need a new login. If `useRefreshTokensFallback` is enabled and the `Auth0` session still exists, silent auth may still work.

Should we say that a new login is needed only if silent auth also cannot work ?

EXAMPLES.md
```diff
@@ -144,6 +144,64 @@ await createAuth0Client({
...
+
+**Difference from `logout()`:** `revokeRefreshToken()` invalidates the refresh token on the Auth0 server and removes it from the local cache, but it does **not** clear the user's Auth0 session or the rest of the local cache. If you want to fully terminate the session, use `logout()` instead.
+
+#### Error Handling
+
+`revokeRefreshToken()` throws a `GenericError` if the `/oauth/revoke` endpoint returns an error (for example, if the token has already been revoked or is invalid). Wrap the call in a try/catch:
```
nandan-bhat: In the worker path, revoke errors come back as a plain `Error`, and that is the default path when refresh tokens use memory cache.

Should we avoid saying this always throws `GenericError` or change the example to catch `Error` or explain the difference  ?

EXAMPLES.md
yogeshchoudhary147: Good catch — updated to mention the silent auth fallback via iframe when `useRefreshTokensFallback` is enabled.

EXAMPLES.md
yogeshchoudhary147: Valid — worker path rejects with a plain `Error`; stripped the `GenericError` reference for now and will revisit once the error type is normalised across both paths.

nandan-bhat (approved):

PR #1595, opened 2026-04-13
feat: add configuration option to opt out of is-authenticated cookies

lorenzolosa:
This library sets a cookie named `auth0.<CLIENT_ID>.is.authenticated` (and optionally a legacy version of the same cookie) to optimize the authentication calls. While this is an effective optimization, it requires setting a cookie that would not be otherwise be needed. In situations where usage of cookies is restricted (or subject to opt-in), for instance for compliance reasons, this might be undesirable.

This PR introduces an option to disable these cookies. This is optional and the default behaviour is not changed: cookies are set unless explicitly disabled. The change is backward-compatible.

### Changes

This PR adds an optional client configuration flag, `useIsAuthenticatedCookies`, on `Auth0ClientOptions` (documented in `src/global.ts`). It defaults to `true`, preserving existing behavior.

When set to false:
* `checkSession` no longer bails out when the `auth0.<clientId>.is.authenticated` cookie (or the legacy `_legacy_` variant) is missing. The SDK still performs the session check (e.g. iframe flow) so silent auth can work in setups that avoid those cookies.
* After successful auth, the SDK does not persist the "is authenticated" cookie via `cookieStorage.save` for that client, which helps in environments where cookies are restricted or undesirable.

API surface: one new optional property on `Auth0ClientOptions`; no removals or deprecations. No UI changes.

### Testing

New unit tests have been added to check that, if `useIsAuthenticatedCookies` is `false`, then:
* `checkSession()` checks the session regardless of whether cookies are set
* `getTokenSilently()` does not set cookies
* `loginWithPopup()` does not set cookies
* `loginWithRedirect()` does not set cookies

### Checklist


yogeshchoudhary147:
Hey, sorry about my earlier comment, that was off track\! The implementation looks good. Before we move forward, could you help us understand what specific compliance requirement you are trying to meet? That will help us understand if this change fully solves your problem.

yogeshchoudhary147:
Closing this PR due to no activity following our request for clarification.

As noted in my earlier comment, we need to understand the specific compliance requirement this addresses before we can move forward. The feature itself may be valid, but without that context we cannot evaluate whether the implementation fully solves the problem or whether there are edge cases we need to account for.

If you have that information available or want to revisit this, feel free to reopen or submit a new PR with the compliance context included. Thank you for the contribution.

PR #1605, opened 2026-05-11
fix: ensure DPoP storage is fully cleared before logout completes

OWConnoi:
### What

This updates `DpopStorage.deleteBy()` to wait for all matching IndexedDB delete requests before resolving.

### Why

`Dpop.clear()` is awaited during logout and calls both `clearNonces()` and `clearKeyPairs()`. Before this change, `deleteBy()` started matching deletes but did not await them, so logout cleanup could continue before the DPoP nonce/keypair delete requests had settled.

This is a small defense-in-depth fix that makes the storage cleanup contract match the caller's expectations.

### Tests

- `npm test -- --runInBand __tests__/dpop/storage.test.ts`

Additional local verification: `npm test -- --runInBand __tests__/dpop/storage.test.ts __tests__/Auth0Client/logout.test.ts` printed PASS for both suites before Jest reported an existing open-handle warning and did not exit promptly.

### DIFF
--- a/__tests__/Auth0Client/logout.test.ts
+++ b/__tests__/Auth0Client/logout.test.ts
@@ -234,6 +234,21 @@ describe('Auth0Client', () => {
       expect(dpop.clear).toHaveBeenCalled();
     });
 
+    it('continues logout when DPoP clear fails', async () => {
+      const auth0 = setup({ useDpop: true });
+      const dpop = auth0['dpop']!;
+
+      jest
+        .spyOn(dpop, 'clear')
+        .mockRejectedValue(new Error('dpop storage unavailable'));
+
+      await expect(auth0.logout()).resolves.not.toThrow();
+
+      expect(window.location.assign).toHaveBeenCalledWith(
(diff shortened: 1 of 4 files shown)


yogeshchoudhary147 (commented):
thanks for the fix! `dpop.clear()` can now throw on failure, which would skip the logout redirect, worth wrapping in a `try/catch`.

OWConnoi:
Thanks, handled. I wrapped DPoP clear during logout so storage cleanup failures do not block the logout redirect, and added a regression test for that path.\n\nVerified with:\n- npx jest __tests__/dpop/storage.test.ts --runInBand\n- npx jest __tests__/Auth0Client/logout.test.ts --runInBand (all tests passed; Jest then reported an existing open-handle warning)\n- npx jest __tests__/Auth0Client/logout.test.ts -t 'continues logout when DPoP clear fails' --runInBand --forceExit

yogeshchoudhary147:
Commits need to be GPG-signed before we can merge. Can you get those signed?

OWConnoi:
The branch has been updated to a GPG-signed squash commit. GitHub is showing the current head commit as verified now.\n\nCurrent head: 1820844ec884c71280c58c21401945237a2174c2

yogeshchoudhary147 (approved):

PR #1608, opened 2026-05-20
feat: CTE delegation and impersonation support

yogeshchoudhary147:
## Summary

Adds delegation and impersonation support for Custom Token Exchange (CTE) per RFC 8693, allowing an actor (e.g. an AI agent or support rep) to act on behalf of a user.

- Add `customTokenExchange()` for stateless token exchange (no session side effects)
- Add `actor_token` and `actor_token_type` to `CustomTokenExchangeOptions`
- Add `ActClaim` type and `act` claim to `IdToken` and `User`
- Add `skipTokenStorage` flag so `refresh_token` is discarded inside the worker and never reaches the main thread (non-worker path returns the raw response as documented)
- `customTokenExchange()` validates `id_token` when present, consistent with `loginWithCustomTokenExchange`

## Test plan

- [x] All existing tests pass (938 passing)
- [x] `customTokenExchange()`: no session side effects, correct params, actor token, audience/org fallback
- [x] `customTokenExchange()`: `id_token` verified when present, skipped when absent
- [x] Worker `skipTokenStorage`: response stripping, no storage side-effect, no overwrite of existing tokens
- [x] `loginWithCustomTokenExchange()` with actor token, `isAuthenticated()` returns true
- [x] `getIdTokenClaims().act` and `getUser().act` populated correctly
- [x] No error when `refresh_token` is absent (suppressed by Auth0 when `actor_token` is present)
- [x] Auth0 error codes surfaced via standard SDK error types

## Intentional omissions

- **Paired parameter validation:** Auth0 returns a clear error (`invalid_request`). No benefit duplicating this client-side.
- **URI validation:** Auth0 validates server-side and returns a clear error. Duplicating adds complexity with no real benefit.

### DIFF
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -1813,7 +1813,7 @@ export class Auth0Client {
   async customTokenExchange(
     options: CustomTokenExchangeOptions
   ): Promise<TokenEndpointResponse> {
-    return oauthToken(
+    const result = await oauthToken(
       {
         baseUrl: this.domainUrl,
         client_id: this.options.clientId,
@@ -1826,6 +1826,12 @@ export class Auth0Client {
       this.worker,
       true // skipTokenStorage — when using a worker, refresh_token is discarded inside it
     );
+
+    if (result.id_token) {
+      await this._verifyIdToken(result.id_token, undefined, options.organization);
+    }
+
+    return result;
   }
 
   /**
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -1750,14 +1750,16 @@ export class Auth0Client {
    * }
    * ```
    */
-  async loginWithCustomTokenExchange(
+  private _buildTokenExchangeParams(
     options: CustomTokenExchangeOptions
-  ): Promise<TokenEndpointResponse> {
-    return this._requestToken({
+  ): TokenExchangeRequestOptions {
+    return {
       ...options,
       grant_type: 'urn:ietf:params:oauth:grant-type:token-exchange',
       subject_token: options.subject_token,
       subject_token_type: options.subject_token_type,
+      ...(options.actor_token && { actor_token: options.actor_token }),
+      ...(options.actor_token_type && { actor_token_type: options.actor_token_type }),
       scope: scopesToRequest(
         this.scope,
         options.scope,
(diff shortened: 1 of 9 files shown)


src/Auth0Client.ts
```diff
@@ -1750,22 +1750,82 @@ export class Auth0Client {
...
+   * ```
+   */
+  async customTokenExchange(
+    options: CustomTokenExchangeOptions
+  ): Promise<TokenEndpointResponse> {
+    return oauthToken(
+      {
+        baseUrl: this.domainUrl,
+        client_id: this.options.clientId,
+        auth0Client: this.options.auth0Client,
```
kishore7snehil, line 1821: When no Web Worker is configured, `customTokenExchange()` calls `oauthToken()` with `skipTokenStorage: true`, but the non-worker path (`fetchWithoutWorker`) does not use `skipTokenStorage` - it returns the raw JSON response including any `refresh_token`. The worker path always strips it, so this is inconsistent. The caller receives the refresh_token in the non-worker path which could lead to accidental token leakage.

Can we strip `refresh_token` from the response in `customTokenExchange()` before returning, regardless of whether a worker is used. Something like this :
```typescript
const { refresh_token, ...safeResult } = await oauthToken(...);
return safeResult;
```

src/Auth0Client.ts
```diff
@@ -1750,22 +1750,82 @@ export class Auth0Client {
...
+        timeout: this.httpTimeoutMs,
+        dpop: this.dpop,
+        ...this._buildTokenExchangeParams(options)
+      },
+      this.worker,
+      true // skipTokenStorage — when using a worker, refresh_token is discarded inside it
+    );
   }
 
   /**
```
kishore7snehil, line 1837: The new `customTokenExchange()` method returns the `oauthToken()` response directly without calling `_verifyIdToken()`, unlike `_requestToken()`. With a fixed Auth0 domain over TLS this has very high prerequisites for exploitation, but applications using the returned `id_token` for authorization decisions would be accepting unvalidated identity data.

We can consider validating the ID token before returning (without writing to cache/cookies) to maintain defense-in-depth.

src/Auth0Client.ts
```diff
@@ -1750,22 +1750,82 @@ export class Auth0Client {
...
+        client_id: this.options.clientId,
+        auth0Client: this.options.auth0Client,
+        useFormData: this.options.useFormData,
+        timeout: this.httpTimeoutMs,
+        dpop: this.dpop,
+        ...this._buildTokenExchangeParams(options)
+      },
+      this.worker,
+      true // skipTokenStorage — when using a worker, refresh_token is discarded inside it
+    );
```
kishore7snehil, line 1828: Because `CustomTokenExchangeOptions` has an index signature (`[key: string]: unknown`), user-controlled options could theoretically include keys like `baseUrl` that override internal transport fields when spread into the `oauthToken` call. Attack prerequisites are high (requires developer to pass untrusted input directly as options), but the defensive fix is trivial.

> You try applying fixed transport fields (baseUrl, client_id, auth0Client, useFormData, timeout) after the user options spread to prevent overrides.

src/worker/token.worker.ts
```diff
@@ -168,24 +168,29 @@ const messageHandler = async ({
...
-    if (json.refresh_token) {
-      // If useMrrt is configured to true we want to save the latest refresh_token
-      // to be used when refreshing tokens with MRRT
-      if (useMrrt) {
-        refreshTokens["latest_refresh_token"] = json.refresh_token;
+    if (!skipTokenStorage) {
+      if (json.refresh_token) {
+        // If useMrrt is configured to true we want to save the latest refresh_token
+        // to be used when refreshing tokens with MRRT
+        if (useMrrt) {
```
kishore7snehil: The `skipTokenStorage` branching correctly prevents both storing and deleting refresh tokens. However, the logic is interleaved with the normal flow rather than being an explicit early-exit path, making it fragile to future edits.

You can handle `skipTokenStorage` as an early branch: after parsing json, if `skipTokenStorage` is true, `delete json.refresh_token` and immediately `postMessage` without any refresh-token store mutation.

kishore7snehil (changes requested):

src/Auth0Client.ts
yogeshchoudhary147: The worker strips `refresh_token` because of its own internal security model (unconditional for all calls), not as a contract of `customTokenExchange`. Stripping on the non-worker path would be silent data loss against an explicit `TokenEndpointResponse` return type; behavior is intentional and documented.

src/Auth0Client.ts
yogeshchoudhary147: Added `_verifyIdToken` after the `oauthToken` call when `id_token` is present. No cache or session side effects, just validation before returning.

src/Auth0Client.ts
yogeshchoudhary147: Good catch. Spread order flipped so transport fields always win regardless of what keys are in the options object.

src/worker/token.worker.ts
yogeshchoudhary147: Refactored to an early exit for `skipTokenStorage`.

kishore7snehil (approved):
LGTM!

PR #1618, opened 2026-06-02
chore: remove webauthn enroll/verify types from MyAccount

yogeshchoudhary147:
## Summary

- Remove `webauthn-platform` and `webauthn-roaming` from enrollment challenge options, responses, and verify options
- Keep `WebAuthnPlatformAuthenticationMethod` and `WebAuthnRoamingAuthenticationMethod` as GET is supported
- Remove related tests and EXAMPLES.md section

## Test plan

- [x] `npm test` passes
- [x] `npm run build` succeeds


src/myaccount/types.ts
```diff
@@ -65,9 +65,7 @@ export type AuthenticationMethodType =
   | 'totp'
   | 'email'
   | 'push-notification'
-  | 'recovery-code'
-  | 'webauthn-platform'
-  | 'webauthn-roaming';
```
pmathew92, line 70: Get operations are supported. So this shouldn't be deleted

yogeshchoudhary147: good catch! Updated

pmathew92 (approved):

PR #1621, opened 2026-06-08
feat: support resource field in MRRT token responses

yogeshchoudhary147:
## Summary

- Adds `resource?: string` to `TokenEndpointResponse` and `CacheEntry`. The field is returned by the server when exchanging a Multi-Resource Refresh Token (MRRT) and indicates the audience the issued access token is scoped for
- Validates the `resource` field against the requested audience before caching; throws `MrrtResourceMismatchError` if they diverge, preventing a mismatched token from ever being stored
- Propagates `resource` through `getTokenSilently({ detailedResponse: true })`, both from a fresh token fetch and from a cache hit

## Behaviour

- Only active when `useMrrt: true` and the server includes `resource` in the response (MRRT exchanges only); non-MRRT flows are unaffected
- `MrrtResourceMismatchError` (extends `GenericError`, error code `mrrt_resource_mismatch`) carries the `requested` and `received` audiences for debugging

## Test plan

- [ ] `useMrrt: true` + server returns matching `resource` -> token cached and returned normally, `resource` present in verbose response
- [ ] `useMrrt: true` + server returns mismatched `resource` -> `MrrtResourceMismatchError` thrown, nothing written to cache
- [ ] `useMrrt: true` + server omits `resource` -> no validation, existing behaviour unchanged
- [ ] `useMrrt: false` -> `resource` field ignored even if server sends it

### DIFF
--- a/__tests__/Auth0Client/mrrt-resource-field.test.ts
+++ b/__tests__/Auth0Client/mrrt-resource-field.test.ts
@@ -1,0 +1,332 @@
+import { verify } from '../../src/jwt';
+import { MessageChannel } from 'worker_threads';
+import * as utils from '../../src/utils';
+
+import {
+  fetchResponse,
+  setupFn,
+  loginWithRedirectFn
+} from './helpers';
+
+import {
+  TEST_ACCESS_TOKEN,
+  TEST_ID_TOKEN,
+  TEST_REFRESH_TOKEN,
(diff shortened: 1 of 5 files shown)


yogeshchoudhary147:
Closing this PR as the `resource` field approach for MRRT has been officially dropped.

PR #1626, opened 2026-06-11
refactor: centralise MFA context storage across all token grant types

yogeshchoudhary147:
## Summary

- Moves MFA context storage out of individual call sites into `_requestToken` so any grant type that receives `mfa_required` from the token endpoint automatically stores MFA context
- Extracts `_storeMfaContext` private helper to avoid duplicating the logic, used by `_requestToken` and `customTokenExchange`
- Adds MFA context storage to `customTokenExchange` which bypasses `_requestToken` by design (no session side effects) but still needs to store context so `mfa.verify()` works
- Fixes scope stored in MFA context to use `scopesToRequest || options.scope` to ensure correctness in the MRRT case where the token endpoint receives a broader cached scope than what was requested
- Fixes `customTokenExchange` to store the resolved scope and audience from `_buildTokenExchangeParams` in MFA context rather than the raw caller values — without this, a caller that omits `audience` would store `undefined` even though the request used `authorizationParams.audience`
- Removes redundant `MfaRequiredError` handling from `MfaApiClient.verify` which was storing MFA context a second time with the same values, now handled centrally by `_requestToken`
- Guards `authorization_code` grant since MFA is handled by Universal Login during the authorization phase, not via a token endpoint error

## Test plan

- [x] `requestTokenForPasskey.test.ts` - passkey grant stores MFA context and re-throws
- [x] `getTokenSilently.test.ts` - refresh token grant stores MFA context and re-throws
- [x] `handleRedirectCallback.test.ts` - `authorization_code` grant does not store MFA context
- [x] `exchangeToken.test.ts` - `loginWithCustomTokenExchange` stores MFA context end-to-end via real fetch mock (previously only tested re-throw via `_requestToken` mock)
- [x] `exchangeToken.test.ts` - `customTokenExchange` stores MFA context and re-throws
- [x] `PasskeyApiClient.test.ts` - `MfaRequiredError` is re-thrown from signup/login

### DIFF
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -1635,6 +1635,8 @@ export class Auth0Client {
     }
   }
 
+  // Central MFA context store — called from _requestToken for all non-authorization_code
+  // grants (including via _requestTokenForMfa) and from customTokenExchange.
   private _storeMfaContext(e: unknown, scope?: string, audience?: string): void {
     if (e instanceof MfaRequiredError) {
       this.mfa.setMFAAuthDetails(e.mfa_token, scope, audience, e.mfa_requirements);
--- a/src/Auth0Client.ts
+++ b/src/Auth0Client.ts
@@ -1548,15 +1548,6 @@ export class Auth0Client {
         }
       }
 
-      if (e instanceof MfaRequiredError) {
-        this.mfa.setMFAAuthDetails(
-          e.mfa_token,
-          options.authorizationParams?.scope,
-          options.authorizationParams?.audience,
-          e.mfa_requirements
-        );
-      }
-
       throw e;
     }
   }
@@ -1644,6 +1635,12 @@ export class Auth0Client {
     }
   }
 
+  private _storeMfaContext(e: unknown, scope?: string, audience?: string): void {
+    if (e instanceof MfaRequiredError) {
+      this.mfa.setMFAAuthDetails(e.mfa_token, scope, audience, e.mfa_requirements);
+    }
+  }
+
   private async _requestToken(
     options:
       | PKCERequestTokenOptions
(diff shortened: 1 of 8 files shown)


src/Auth0Client.ts
```diff
@@ -1644,6 +1635,12 @@ export class Auth0Client {
     }
   }
 
+  private _storeMfaContext(e: unknown, scope?: string, audience?: string): void {
```
Piyush-85, line 1640: This chain — verify() → _requestTokenForMfa() → _requestToken() → _storeMfaContext() — is undocumented. A future reader looking at MfaApiClient.verify() has no indication that sequential-factor mfa_required is handled. 
Please add a comment or proper documentation.

src/Auth0Client.ts
Piyush-85, line 1: `mfa.getAuthenticators()` throws for global MFA policy

This PR makes passkey+MFA and CTE+MFA reachable for the first time, but `getAuthenticators` immediately throws for any tenant using a global MFA policy:

```typescript
if (!context?.mfaRequirements?.challenge || context.mfaRequirements.challenge.length === 0) {
  throw new MfaListAuthenticatorsError('invalid_request', 'challengeType is required...');
}
```

From the Auth0 server code, `mfa_requirements.challenge` has exactly two states:

| Trigger | `mfa_requirements.challenge` |
|---|---|
| Global MFA policy | absent — `mfa_requirements` omitted entirely |
| Step-up / assurance policy | present, always non-empty |

For global MFA policy, `context.mfaRequirements?.challenge` is `undefined` — the guard throws with a misleading error and the MFA flow is dead immediately after `MfaRequiredError` is caught. The use case this PR enables is blocked for the most common MFA configuration.

yogeshchoudhary147: Good catch. Confirmed via testing passkey login on a tenant with global MFA enabled, response has no `mfa_requirements` and the guard was throwing immediately.

Fixed! `getAuthenticators` now returns all authenticators unfiltered for global MFA policy and filters by challenge types only for step-up.

src/Auth0Client.ts
yogeshchoudhary147: Added comments in both `_storeMfaContext` and `MfaApiClient.verify`.

Piyush-85 (approved):

Piyush-85:
LGTM !

PR #1633, opened 2026-06-19
ci: enforce conventional commits format on PR titles 

yogeshchoudhary147:
## Summary

- Adds a `lint-pr-title` local action to validate PR titles against the Conventional Commits specification using commitlint
- Adds `commitlint.config.mjs` at the repo root with inline rules (no external `extends` dependency)
- Adds `.github/workflows/lint-pr-title.yml` triggered on `pull_request` (works for forks on public repos without needing elevated permissions)
- Action runs on Node 24 with ESM modules

## Test plan

- [x] Open a PR with an invalid title (e.g. `my change`) and verify the check fails
- [x] Open a PR with a valid title (e.g. `fix: correct token expiry edge case`) and verify the check passes
- [ ] Verify the check runs on PRs from forks

### DIFF
--- a/.github/actions/lint-pr-title/src/index.js
+++ b/.github/actions/lint-pr-title/src/index.js
@@ -1,0 +1,63 @@
+import { default as load } from '@commitlint/load';
+import { default as lint } from '@commitlint/lint';
+import { default as format } from '@commitlint/format';
+import * as core from '@actions/core';
+import * as github from '@actions/github';
+
+async function run() {
+  try {
+    // 1. Get the pull request title
+    const prTitle = github.context.payload.pull_request?.title;
+
+    if (!prTitle) {
+      core.setFailed('⛔️ Could not get pull request title. This action should only be run on a pull_request event.');
+      return;
+    }
+
+    core.info(`📝 Validating PR Title: "${prTitle}"`);
+
+    // 2. Load the commitlint configuration from the repository
+    const config = await load();
+    core.info('✅ Loaded commitlint configuration successfully.');
+
+    // 3. Lint the pull request title
+    const result = await lint(prTitle, config.rules, {
+      defaultIgnores: config.defaultIgnores,
+      helpUrl: config.helpUrl,
+    });
+
+    // 4. Report the results
+    if (result.valid) {
+      core.info('✅ PR title is valid.');
+    } else {
+      const errors = format({ results: [result] }, { color: false }).trimEnd();
+      const validTypes = config.rules['type-enum']?.[2]?.join(', ');
+      const sep = '─'.repeat(49);
+      const details = [
+        errors,
+        '',
+        sep,
+        ' Expected format:  <type>(<optional scope>): <subject>',
+        ...(validTypes ? ['', ` Valid types:  ${validTypes}`] : []),
+        '',
+        ' Examples:',
+        '   feat: add login with passkeys',
+        '   fix(cache): correct token expiry edge case',
+        '   ci: add PR title linting',
+        '',
+        ' Learn more: https://www.conventionalcommits.org',
+        sep,
+      ].join('\n');
+      core.info(details);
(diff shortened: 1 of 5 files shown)


.github/actions/lint-pr-title/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!prTitle) {
+      core.setFailed('⛔️ Could not get pull request title. This action should only be run on a pull_request event.');
+      return;
+    }
+
+    core.info(`📝 Validating PR Title: "${prTitle}"`);
+
+    // 2. Load the commitlint configuration from the repository
+    const config = await load();
```
arpit-jn: `@commitlint/load` does not throw when no config file is discovered. Tt returns `{ rules: {} }`. With empty rules, `lint(prTitle, {})` returns `valid: true` for every title, while the action still logs "Loaded successfully." So if `commitlint.config.mjs` is ever renamed, moved, or the action's cwd shifts, enforcement is silently disabled, and all PRs pass, the worst failure mode for a guard rail. Could we add an assertion?

```
const config = await load();
if (!config.rules || Object.keys(config.rules).length === 0) {
  core.setFailed('⛔️ No commitlint rules loaded. Is commitlint.config.mjs present at the repo root?');
  return;
}
```

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Checkout code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install action dependencies
+        run: npm install
```
arpit-jn: On a `pull_request` event, the fork's head is checked out, so its `package.json` is attacker-controlled; `npm install` runs `postinstall`, etc. → arbitrary code on the runner. Blast radius is limited (read-only token, no secrets on fork PRs, but `--ignore-scripts` is free hardening and is fully compatible with your no-lockfile/float-to-patched design:

`run: npm install --ignore-scripts`

This is exactly the path your unchecked test-plan box ("runs on PRs from forks") exercises.

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,37 @@
+name: Lint PR Title
+
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+
+permissions:
```
arpit-jn, line 9: The title is read from `github.context.payload` (the event), never via API, so `pull-requests: read` is unused, while `actions/checkout` wants `contents: read`. It works today only because anonymous read works on a public repo. (Dropping my earlier "repo goes private" justification, it's public and staying so.) Set:

```
permissions:
  contents: read
```

.github/actions/lint-pr-title/src/index.js
yogeshchoudhary147: Fixed — passing explicit `file` to `load()` so it throws on missing config, plus a rules-length guard as a second layer.

.github/workflows/lint-pr-title.yml
yogeshchoudhary147: Fixed — added `--ignore-scripts`; also noted the root `.npmrc` has `ignore-scripts=true` but doesn't apply here since the action has its own `package.json`.

.github/workflows/lint-pr-title.yml
yogeshchoudhary147: Fixed — `pull-requests: read` → `contents: read`, consistent with every other workflow in this repo that only needs checkout.

arpit-jn (approved):

PR #1634, opened 2026-06-20
feat: enforce IPSIE session_expiry ceiling on local session lifetime

yogeshchoudhary147:
## Summary

Implements IPSIE SL1 `session_expiry` enforcement. When an ID token carries a `session_expiry` Unix-timestamp claim, the SDK treats it as a hard ceiling on the local session. No token is served and no refresh is attempted once the ceiling has passed.

## Behaviour

| Scenario | Behaviour |
|---|---|
| `session_expiry` absent | No change, fully backward compatible |
| Ceiling not yet reached | Tokens served normally |
| Within 30s leeway window | Treated as expired (clock-skew guard) |
| Ceiling passed | `getUser` / `getIdTokenClaims` / `getTokenSilently` return `undefined`; local session fully torn down |
| `session_expiry <= iat` at login | Rejected immediately with `GenericError('invalid_token')`, session not persisted |
| Token refresh | `/oauth/token` not called past ceiling; ceiling pinned to login-time value |
| Re-login without logout (authorization_code) | Fresh login token claims are authoritative; old ceiling is not inherited |

## Tests

24 new unit tests in `__tests__/Auth0Client/session-expiry.test.ts` covering:
- Login-time validation via `loginWithRedirect` and `loginWithPopup`
- Ceiling enforcement across `getTokenSilently`, `getUser`, `isAuthenticated`, `getIdTokenClaims`
- Leeway boundary, no network call on breach
- Ceiling pinning through silent refresh (refresh token grant) only — fresh `authorization_code` logins do not inherit a prior ceiling
- Full local teardown on breach (cache, cookies, DPoP keys, worker)

## Testing

Deploy the following Post-Login Action on your tenant (Dashboard → Actions → Triggers → post-login):

```js
exports.onExecutePostLogin = async (event, api) => {
  api.idToken.setCustomClaim('session_expiry', Math.floor(Date.now() / 1000) + 120);
};
```

Verified: tokens served normally before ceiling, all session-read methods return `undefined` after breach, no `/oauth/token` call on breach, ceiling not extended by forced refresh.

### DIFF
--- a/__tests__/Auth0Client/session-expiry.test.ts
+++ b/__tests__/Auth0Client/session-expiry.test.ts
@@ -1,0 +1,507 @@
+import { verify } from '../../src/jwt';
+import { MessageChannel } from 'worker_threads';
+import * as utils from '../../src/utils';
+import * as scope from '../../src/scope';
+import { expect } from '@jest/globals';
+
+import * as esCookie from 'es-cookie';
+// @ts-ignore — resolved to the test mock via jest.mock() below
+import TokenWorker from '../../src/worker/token.worker';
+
+import { GenericError } from '../../src/errors';
+import {
+  fetchResponse,
+  loginWithPopupFn,
+  loginWithRedirectFn,
+  setupFn
+} from './helpers';
+import {
+  TEST_ACCESS_TOKEN,
+  TEST_CLIENT_ID,
+  TEST_CODE_CHALLENGE,
+  TEST_ID_TOKEN,
+  TEST_REFRESH_TOKEN,
+  TEST_TOKEN_TYPE,
+  nowSeconds
+} from '../constants';
+
+jest.mock('es-cookie');
+jest.mock('../../src/jwt');
+jest.mock('../../src/worker/token.worker');
+
+const mockWindow = <any>global;
+const mockFetch = <jest.Mock>mockWindow.fetch;
+const mockVerify = <jest.Mock>verify;
+
+jest
+  .spyOn(utils, 'bufferToBase64UrlEncoded')
+  .mockReturnValue(TEST_CODE_CHALLENGE);
+
+const setup = setupFn(mockVerify);
+const loginWithRedirect = loginWithRedirectFn(mockWindow, mockFetch);
+const loginWithPopup = loginWithPopupFn(mockWindow, mockFetch);
+
+describe('Auth0Client', () => {
+  const oldWindowLocation = window.location;
+
+  beforeEach(() => {
+    delete window.location;
+    window.location = Object.defineProperties(
+      {},
+      {
+        ...Object.getOwnPropertyDescriptors(oldWindowLocation),
+        assign: {
+          configurable: true,
+          value: jest.fn()
+        }
+      }
+    ) as Location;
+
+    mockWindow.open = jest.fn();
+    mockWindow.addEventListener = jest.fn();
+    mockWindow.removeEventListener = jest.fn();
+
+    mockWindow.crypto = {
+      subtle: { digest: () => 'foo' },
+      getRandomValues() {
+        return '123';
+      }
+    };
+
+    mockWindow.MessageChannel = MessageChannel;
+    mockWindow.Worker = {};
+
+    jest.spyOn(scope, 'getUniqueScopes');
+    sessionStorage.clear();
+  });
+
+  afterEach(() => {
+    mockFetch.mockReset();
+    jest.clearAllMocks();
+    window.location = oldWindowLocation;
+  });
+
+  describe('session_expiry', () => {
+    describe('login-time validation', () => {
+      it('throws if session_expiry equals iat', async () => {
+        const iat = nowSeconds();
+        const auth0 = setup({}, { iat, session_expiry: iat });
+
+        const error = await loginWithRedirect(auth0).catch(e => e);
+        expect(error).toBeInstanceOf(GenericError);
+        expect(error.error).toBe('invalid_token');
+        expect(error.message).toBe(
+          'Invalid session_expiry: session ceiling is before or at the token issue time.'
+        );
+      });
+
+      it('throws if session_expiry is before iat', async () => {
+        const iat = nowSeconds();
+        const auth0 = setup({}, { iat, session_expiry: iat - 1 });
+
+        const error = await loginWithRedirect(auth0).catch(e => e);
+        expect(error).toBeInstanceOf(GenericError);
+        expect(error.error).toBe('invalid_token');
+        expect(error.message).toBe(
+          'Invalid session_expiry: session ceiling is before or at the token issue time.'
+        );
+      });
+
+      it('succeeds if session_expiry is after iat', async () => {
+        const iat = nowSeconds();
+        const auth0 = setup({}, { iat, session_expiry: iat + 3600 });
+
+        await expect(loginWithRedirect(auth0)).resolves.not.toThrow();
+      });
+
+      it('succeeds when session_expiry is absent', async () => {
+        const auth0 = setup();
+
+        await expect(loginWithRedirect(auth0)).resolves.not.toThrow();
+      });
+    });
+
+    describe('getTokenSilently ceiling enforcement', () => {
+      it('returns a token when no session_expiry claim is present', async () => {
+        const auth0 = setup();
+        await loginWithRedirect(auth0);
+        mockFetch.mockReset();
+
+        const token = await auth0.getTokenSilently();
+
+        expect(token).toBeTruthy();
+      });
+
+      it('returns a token when the ceiling has not been reached', async () => {
+        const iat = nowSeconds();
+        const sessionExpiry = iat + 3600;
+        const auth0 = setup(
+          { nowProvider: () => (sessionExpiry - 100) * 1000 },
+          { iat, session_expiry: sessionExpiry }
+        );
+        await loginWithRedirect(auth0);
+        mockFetch.mockReset();
+
+        const token = await auth0.getTokenSilently();
+
+        expect(token).toBeTruthy();
+      });
+
+      it('returns undefined when the ceiling has been breached', async () => {
+        const iat = nowSeconds();
+        const sessionExpiry = iat + 3600;
+        const auth0 = setup(
+          { nowProvider: () => (sessionExpiry + 100) * 1000 },
+          { iat, session_expiry: sessionExpiry }
+        );
+        await loginWithRedirect(auth0);
+        mockFetch.mockReset();
+
+        const token = await auth0.getTokenSilently();
+
+        expect(token).toBeUndefined();
+      });
+
+      it('returns undefined when within the 30s leeway window', async () => {
+        const iat = nowSeconds();
+        const sessionExpiry = iat + 3600;
+        // nowSeconds = sessionExpiry - 15, which is >= sessionExpiry - 30
+        const auth0 = setup(
+          { nowProvider: () => (sessionExpiry - 15) * 1000 },
+          { iat, session_expiry: sessionExpiry }
+        );
+        await loginWithRedirect(auth0);
+        mockFetch.mockReset();
+
+        const token = await auth0.getTokenSilently();
+
+        expect(token).toBeUndefined();
+      });
+
+      it('returns a token when just outside the 30s leeway window', async () => {
+        const iat = nowSeconds();
+        const sessionExpiry = iat + 3600;
+        // nowSeconds = sessionExpiry - 31, which is < sessionExpiry - 30
+        const auth0 = setup(
+          { nowProvider: () => (sessionExpiry - 31) * 1000 },
+          { iat, session_expiry: sessionExpiry }
+        );
+        await loginWithRedirect(auth0);
+        mockFetch.mockReset();
+
+        const token = await auth0.getTokenSilently();
+
+        expect(token).toBeTruthy();
+      });
+
+      it('does not reach the network on breach', async () => {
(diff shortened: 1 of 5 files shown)


src/Auth0Client.ts
```diff
@@ -1571,6 +1564,56 @@ export class Auth0Client {
...
+  private async _isSessionCeilingReached(): Promise<boolean> {
+    const inMemory = this.userCache.get<IdTokenEntry>(
+      CACHE_KEY_ID_TOKEN_SUFFIX
+    ) as IdTokenEntry;
+    const idTokenEntry =
+      inMemory ??
+      await this.cacheManager.getIdToken(
+        new CacheKey({ clientId: this.options.clientId })
+      );
+    const sessionExpiresAt = idTokenEntry?.decodedToken?.claims?.session_expiry;
```
kishore7snehil, line 1619: `session_expiry` is read straight off the decoded claims here and compared as-is, with nothing checking it's a sane value first. Since the claim is set by a customer Post-Login Action rather than the platform, it's effectively untrusted input, and the comparison reacts differently depending on what shape it arrives in. A millisecond value  - someone drops the `/1000` in their Action - reads as a ceiling tens of thousands of years out, so enforcement is silently off while everything looks fine. A zero, negative, or float value goes the other way and is treated as already-expired, which clears the session and locks the user out. A string slips through as "no ceiling" only by accident of `NaN` comparisons. So the same kind of bad input fails open in one case and locks out in another.

Could we run the value through one small helper, used at both this read and the login check, that accepts only a positive integer and treats anything `>= 10_000_000_000` as "no ceiling"?

src/Auth0Client.ts
```diff
@@ -1555,6 +1540,14 @@ export class Auth0Client {
   private async _saveEntryInCache(
     entry: CacheEntry & { id_token: string; decodedToken: DecodedToken }
   ) {
+    const { session_expiry, iat } = entry.decodedToken.claims;
+    if (session_expiry !== undefined && (iat === undefined || session_expiry <= iat)) {
```
kishore7snehil: The login guard here rejects only when `session_expiry <= iat`, with no leeway, but the read-time check expires the session at `session_expiry - 30s`. So a fresh login whose ceiling lands 1–30s in the future passes this guard and gets persisted, then the very next read immediately tears the session down - the user "logs in" and is bounced on the next call, which is the born-dead session this guard is meant to prevent.

The silent-refresh block overwrites `session_expiry` with the pinned value before this runs, so on a refresh the guard ends up validating the pinned ceiling rather than what actually came back. Routing the login guard through the same check the reads use - and validating the incoming claim rather than the pinned one - would keep both ends in agreement. 

A boundary test around the leeway window can be added

src/Auth0Client.ts
```diff
@@ -1687,6 +1730,19 @@ export class Auth0Client {
         }
       }
 
+      // Silent refresh — pin the session_expiry ceiling from the initial login.
+      // Prevents the server from extending the ceiling via a refresh token response.
+      if (options.grant_type !== 'authorization_code') {
```
kishore7snehil, line 1749: This pins the previous ceiling onto everything that isn't an `authorization_code` grant, but that's broader than just silent refresh. `loginWithCustomTokenExchange` runs through `_requestToken` with the token-exchange grant, and the passkey and MFA login paths do too - those are fresh logins that should establish their own ceiling, but here they'll inherit whatever ceiling was cached from a prior session instead.

 The behavior table treats a fresh login as authoritative; this path quietly breaks that for the non-redirect login flows.

Could we narrow this to the actual silent-refresh case - `options.grant_type === 'refresh_token'` - or gate it on an explicit "this is a refresh, not a login" signal? 

We should add test on the token-exchange / passkey login paths to document which way it should go.

src/Auth0Client.ts
```diff
@@ -939,6 +948,10 @@ export class Auth0Client {
   ): Promise<undefined | GetTokenSilentlyVerboseResponse> {
     const { cacheMode, ...getTokenOptions } = options;
 
+    if (await this._isSessionCeilingReached()) {
```
kishore7snehil, line 951: When the ceiling is reached `getTokenSilently` returns `undefined`, but the public method is typed to return a string (or the verbose response), never `undefined`. Callers leaning on the type won't know to handle it, so this surfaces downstream as `Bearer undefined` or a destructuring failure on the verbose shape rather than a clear "you need to log in again."

The same "ceiling reached" condition shows up three different ways - `getTokenSilently` returns `undefined`, the read methods return `undefined`, and login throws `GenericError('invalid_token')`. The silent return on the reads is the right call (it lets the existing redirect-to-login path run untouched), but for `getTokenSilently` either the return type should admit `undefined` or it should throw a typed error the caller can catch. Which way do we want this to read?

src/Auth0Client.ts
```diff
@@ -1571,6 +1564,56 @@ export class Auth0Client {
...
+      inMemory ??
+      await this.cacheManager.getIdToken(
+        new CacheKey({ clientId: this.options.clientId })
+      );
+    const sessionExpiresAt = idTokenEntry?.decodedToken?.claims?.session_expiry;
+    if (sessionExpiresAt === undefined) return false;
+
+    const now = await this.nowProvider();
+    const nowSeconds = Math.floor(now / 1000);
+    if (nowSeconds >= sessionExpiresAt - SESSION_EXPIRY_LEEWAY_SECONDS) {
```
kishore7snehil, line 1624: Once the ceiling is hit, every subsequent `getUser` / `getIdTokenClaims` / `getTokenSilently` call re-runs the full teardown here - clearing already-empty caches and cookies and re-posting a `clear` to the worker - because there's no guard for "already torn down." 

Low impact, but the repeated `sendMessage` to a worker that's already been cleared could surface as flaky. A simple "already cleared" short-circuit would avoid the repeat work.

src/Auth0Client.ts
```diff
@@ -592,6 +593,10 @@ export class Auth0Client {
    * @typeparam TUser The type to return, has to extend {@link User}.
    */
   public async getUser<TUser extends User>(): Promise<TUser | undefined> {
+    if (await this._isSessionCeilingReached()) {
```
kishore7snehil, line 596: This adds real user-facing behavior - a previously-valid session can now start returning `undefined` from `getUser` / `getTokenSilently` once the upstream ceiling is reached - but there's no documentation for it in the PR. The requirements doc lists a docs section as a must-have, so a short "Session expiry from upstream IdP" example would be worth adding: what the claim is, the connection-option prerequisite, how the value is emitted (the Action snippet, with the milliseconds caveat), and an upgrade note that these methods can now return `undefined` for a user who was logged in, so apps that assumed a value should add a null check.

The auth0-server-js examples section for this same feature is a good one to model it on : https://github.com/auth0/auth0-auth-js/blob/cef72915050abe34a921f8fa1566774c76f94511/packages/auth0-server-js/EXAMPLES.md#session-expiry-from-upstream-idp-ipsie-session_expiry

__tests__/Auth0Client/session-expiry.test.ts
```diff
@@ -0,0 +1,507 @@
...
+  });
+
+  afterEach(() => {
+    mockFetch.mockReset();
+    jest.clearAllMocks();
+    window.location = oldWindowLocation;
+  });
+
+  describe('session_expiry', () => {
+    describe('login-time validation', () => {
```
kishore7snehil, line 85: The suite covers the happy path, the leeway boundary, and the `<= iat` rejection well, but the malformed-input cases aren't exercised anywhere. 

There's no test for a millisecond-magnitude value, or for a string / zero / negative / float, each asserting it's treated as "no ceiling" rather than silently disabling enforcement or locking the user out.

Could we add those alongside the validation fix, plus a case for the non-`authorization_code` login paths (token-exchange, passkey, MFA) so the ceiling-pinning behavior on those is pinned down too?

A leeway-boundary case at login would also document what should happen at login versus the next read.

kishore7snehil (changes requested):

src/Auth0Client.ts
yogeshchoudhary147: The born-dead session case will only happen if someone sets the ceiling within 30 seconds of `iat` in their Action. That is basically a misconfigured Action, not a real scenario we need to handle.

On the refresh side, yes the guard sees the pinned value and not what the server actually sent, but we are discarding the server's claim anyway so that is fine.

src/Auth0Client.ts
yogeshchoudhary147: This is intentional. For any fresh login there is no existing session in the cache so `existingCeiling` will be `undefined` and pinning is a no-op regardless of grant type. Keeping it as `grant_type !== 'authorization_code'` means any new refresh-like grant added in future gets ceiling pinning automatically without touching this block.

src/Auth0Client.ts
yogeshchoudhary147: This is a pre-existing gap actually. `cacheMode: 'cache-only'` already returns `undefined` through the same overloads. This PR just follows the same pattern. Can fix the overloads properly in a separate PR.

src/Auth0Client.ts
yogeshchoudhary147: This only runs once actually. After the first teardown the cache is empty, so the next call finds no `session_expiry` and returns `false` immediately without touching anything.

src/Auth0Client.ts
yogeshchoudhary147: Good catch, missed the docs. Will add a section in EXAMPLES.md covering the claim, Action snippet with the ms warning, and an upgrade note for null checks.

src/Auth0Client.ts
yogeshchoudhary147: Millisecond and non-numeric string are valid, will fix both with a throw at login. Zero/negative are already caught by the `<= iat` guard, float is off by 1 second at most so not a real concern. Will throw instead of treating as "no ceiling" since silently disabling enforcement is the same outcome as the bug.

src/Auth0Client.ts
kishore7snehil: The no-op only holds for a first login or one right after logout. loginWithCustomTokenExchange, the passkey path, and MFA all go through _requestToken with a non-authorization_code grant, and none of them clear the cache first ,only the auth-code branch does. So if a user already has a session and then does a CTE / passkey / MFA login, the old cached token is still there, existingCeiling is set, and the stale ceiling gets pinned onto what should be a fresh login. 

Can we gate this on grant_type === 'refresh_token' instead, and add a test on the token-exchange / passkey path?

src/Auth0Client.ts
kishore7snehil: Throwing stops a bad value from silently turning enforcement off, but it has the opposite risk: the claim comes from a customer Action, so one bad value locks every user on that tenant out of login. Treating a malformed claim as "no ceiling" lands in the same place as the claim just being absent, without the lockout. The new tests and the EXAMPLES ms note now bake the throw in.
                                                             
Do you want to fail open or keep the throw?

src/Auth0Client.ts
yogeshchoudhary147: Understood on the cache not being cleared. Quick question though, what is the realistic scenario where CTE or passkey login happens without an explicit logout first? If the user is already authenticated, what drives them to call these flows again mid-session?

src/Auth0Client.ts
yogeshchoudhary147: Keeping the throw. Silent bypass of a security control is worse than a temporary lockout. Developer will notice the login failure immediately and fix the Action, but a silent no ceiling could go unnoticed for weeks.

__tests__/Auth0Client/session-expiry.test.ts
yogeshchoudhary147: Added tests for millisecond and non-numeric string. Zero/negative already covered, float is a non-issue. Non-AC path tests are tied to the comment 3 thread. Sub-30s ceiling is a broken Action, skipping that test.

src/Auth0Client.ts
kishore7snehil: On a scenario without an explicit logout first -  A user is already logged in, then hits something sensitive like a payment, changing security settings, an admin action and the app runs an MFA or passkey check right there to raise assurance. There's no logout in between. it's the same authenticated user proving themselves again, so the prior session (and its cached token) is still sitting there.

On what drives the call mid-session: it's not the user re-logging in, it's the app demanding a higher assurance level for a specific action.So in both cases the cache still holds the previous login's token, existingCeiling is set, and the stale ceiling gets pinned onto the step-up.

yogeshchoudhary147: Step-up is not a fresh login, it is the same session with higher assurance. Pinning the ceiling there is intentional. If step-up could reset it, apps could just trigger an MFA challenge to bypass the IPSIE limit.

Also narrowing to `refresh_token` would leave `mfa-otp` and `mfa-oob` grants unpinned. Server can return a later `session_expiry` on MFA verify and we would store it, defeating the ceiling.

yogeshchoudhary147: Think of it this way. `authorization_code` is the only grant that can establish a ceiling. For everything else:

- **Login grants** (passkey, CTE): no prior session means no existing ceiling, so the new token's `session_expiry` is stored as is.
- **Non-login grants** (refresh_token, step-up with MFA): always mid-session, existing ceiling is always preserved.

This also means any new mid-session grant added in future gets the right behavior automatically without touching this block.

kishore7snehil (approved):
LGTM!

PR #1693, opened 2026-07-20
fix: remove singlePromise from getTokenSilently to respect cacheMode: 'off'

### LINKED ISSUE #1691: cacheMode: 'off' silently receives cached result when an in-flight cacheMode: 'on' call exists
## Checklist

- [x] The issue can be reproduced in the [auth0-spa-js sample app](https://github.com/auth0-samples/auth0-javascript-samples/tree/master/01-Login) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/auth0-spa-js#readme), [Examples](https://github.com/auth0/auth0-spa-js/blob/main/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-spa-js/blob/main/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://auth0.com/docs/libraries/auth0-single-page-app-sdk) and [API documentation](https://auth0.github.io/auth0-spa-js/), and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-spa-js/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

## Description

Calling `getTokenSilently({ cacheMode: 'off' })` concurrently with an in-flight `getTokenSilently()` (default `cacheMode: 'on'`) for the same audience and scope does **not** trigger a network request. Instead, the `cacheMode: 'off'` call silently receives the cached result from the `'on'` call — ignoring the intent to bypass the cache entirely.

**Expected:** a `cacheMode: 'off'` call always makes its own `/oauth/token` request, regardless of any in-flight calls.

### Root cause

`singlePromise` (in `src/promise-utils.ts`) deduplicates calls using the key:

```
`${clientId}::${audience}::${scope}`
```

`cacheMode` is not part of the key, so a `cacheMode: 'off'` call that arrives while a `cacheMode: 'on'` call is in-flight is folded into the existing promise and never invokes `_getTokenSilently` independently.

A possible starting point for a fix: append `::${localOptions.cacheMode}` to the dedup key in `Auth0Client.ts` (the `singlePromise` call inside `getTokenSilently`):

```diff
- `${this.options.clientId}::${localOptions.authorizationParams.audience}::${localOptions.authorizationParams.scope}`
+ `${this.options.clientId}::${localOptions.authorizationParams.audience}::${localOptions.authorizationParams.scope}::${localOptions.cacheMode}`
```

This would prevent a `cacheMode: 'off'` call from being folded into an in-flight `'on'` call, and `'off'` calls would still deduplicate *with each other* (avoiding duplicate concurrent refresh token exchanges).

However, this one-liner has a subtle side effect on `cacheMode: 'cache-only'`: currently a `cache-only` call that races an in-flight `'on'` call is folded into it and gets the fresh token once it resolves; with this change it would run independently and return `undefined` if the cache is empty at that moment. A more correct solution may require a read-write queue — `'on'` and `'cache-only'` as readers that can share a promise, `'off'` as an exclusive writer that waits for in-flight readers and blocks new ones — but we don't have full context on the intended semantics here.

## Reproduction

1. Configure `Auth0Provider` with `useRefreshTokens: true` and `cacheLocation: 'localstorage'`.
2. On app startup, `checkSession()` fires an initial `getTokenSilently()` internally (or any code path that calls it with the default `cacheMode: 'on'`).
3. Before that call resolves, call `getTokenSilently({ cacheMode: 'off' })` — e.g. from a route loader that needs a fresh token with updated scopes.
4. **Observe:** no `/oauth/token` request appears in the network panel. The `'off'` call resolves immediately with the same (potentially stale) result as the `'on'` call. The returned token's `iat` (issued-at) confirms it was not refreshed.

## Additional context

- **auth0/auth0-react#375** — "Access token not reloaded despite `ignoreCache` flag" (2022, still open, no fix). The `setTimeout` workaround described there confirms the race: delaying until the in-flight call settles allows the `ignoreCache`/`cacheMode: 'off'` call to proceed normally. This is the same bug.

The race is most reliably triggered by `Auth0Provider`'s internal `checkSession()` on mount, which holds a `cacheMode: 'on'` promise. Any concurrent `cacheMode: 'off'` call from application code (e.g. a router `beforeLoad`/`loader`) will be silently absorbed into it.

## auth0-spa-js version

2.21.1 (issue present in all versions using `singlePromise`)

## Which framework are you using?

React (`@auth0/auth0-react`)

## Which browsers have you tested in?

Chrome, Safari

yogeshchoudhary147:
## Problem

`getTokenSilently({ cacheMode: 'off' })` was silently receiving the cached result from an in-flight `getTokenSilently()` (default `cacheMode: 'on'`) call for the same audience and scope, instead of making its own `/oauth/token` request.

Root cause: `singlePromise` deduplicated calls using a key of `clientId::audience::scope`. `cacheMode` was not part of the key, so a `cacheMode: 'off'` call arriving while a `cacheMode: 'on'` call was in-flight was folded into the existing promise — `_getTokenSilently` was never invoked for the second caller.

Fixes #1691

## Solution

Remove `singlePromise` from `getTokenSilently` entirely and rely on the existing `lockManager.runWithLock` mechanism to handle concurrency. The lock already serialises concurrent token requests correctly via a double-checked cache pattern:

- First caller acquires the lock, fetches a fresh token, stores it in cache, releases the lock
- Subsequent `cacheMode: 'on'` callers acquire the lock, find the token in cache, return it — no extra network request
- `cacheMode: 'off'` callers acquire the lock, skip the cache read, and fetch their own fresh token — as intended

`singlePromise` remains in `promise-utils.ts` as it may be useful elsewhere.

## Test changes

- Added a regression test for the exact bug scenario: a `cacheMode: 'off'` call concurrent with an in-flight `cacheMode: 'on'` call now correctly triggers two separate network requests
- `_getTokenSilently` is now called once per concurrent caller (not once total) — the lock handles deduplication at the network level
- Cross-instance concurrent calls now each invoke `_getTokenSilently` independently — the shared lock still prevents duplicate network requests
- Two concurrent `cacheMode: 'off'` calls now each correctly make their own request and receive independent fresh tokens
- Renamed two concurrency test descriptions that were reflecting the old `singlePromise` behaviour

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved silent token retrieval behavior for concurrent calls, including more accurate expectations around per-caller work while still reusing compatible in-flight activity.
  * Ensured `cacheMode: 'off'` concurrent calls are not deduplicated and each performs its own token retrieval when required.
  * Updated scenarios involving cache hits and lock-release to reflect correct token and network activity counts.
* **Documentation**
  * Clarified `cacheMode: 'off'` behavior for concurrent `getTokenSilently` calls in option documentation.

### DIFF
--- a/__tests__/Auth0Client/getTokenSilently.test.ts
+++ b/__tests__/Auth0Client/getTokenSilently.test.ts
@@ -4,7 +4,6 @@ import { MessageChannel } from 'worker_threads';
 import * as api from '../../src/api';
 import * as http from '../../src/http';
 import { verify } from '../../src/jwt';
-import * as promiseUtils from '../../src/promise-utils';
 import * as scope from '../../src/scope';
 import * as utils from '../../src/utils';
 
@@ -1031,7 +1030,7 @@ describe('Auth0Client', () => {
         expect(client['_getTokenSilently']).toHaveBeenCalledTimes(2);
       });
 
-      it('should not call _getTokenSilently if a call is already in flight', async () => {
+      it('should call _getTokenSilently for each concurrent caller but only make one network request', async () => {
         const client = setup();
 
         jest.spyOn(<any>utils, 'runIframe').mockResolvedValue({
@@ -1047,11 +1046,11 @@ describe('Auth0Client', () => {
           getTokenSilently(client)
         ]);
 
-        expect(client['_getTokenSilently']).toHaveBeenCalledTimes(1);
+        expect(client['_getTokenSilently']).toHaveBeenCalledTimes(2);
         expect(tokens[0]).toEqual(tokens[1]);
       });
 
-      it('should not call _getTokenSilently if a call is already in flight (cross instance)', async () => {
+      it('should call _getTokenSilently on each instance but only make one network request across instances', async () => {
         const client1 = setup();
         const client2 = setup();
 
@@ -1070,8 +1069,25 @@ describe('Auth0Client', () => {
         ]);
 
         expect(client1['_getTokenSilently']).toHaveBeenCalledTimes(1);
-        expect(client2['_getTokenSilently']).not.toHaveBeenCalled();
+        expect(client2['_getTokenSilently']).toHaveBeenCalledTimes(1);
         expect(tokens[0]).toEqual(tokens[1]);
+      });
+
+      it('should make its own network request when cacheMode is off and a cacheMode on call is in flight', async () => {
+        const client = setup();
+
+        jest.spyOn(<any>utils, 'runIframe').mockResolvedValue({
+          access_token: TEST_ACCESS_TOKEN,
+          state: TEST_STATE,
+          code: TEST_CODE
+        });
+
+        await Promise.all([
+          getTokenSilently(client),
+          getTokenSilently(client, { cacheMode: 'off' })
+        ]);
+
+        expect(utils.runIframe).toHaveBeenCalledTimes(2);
       });
     });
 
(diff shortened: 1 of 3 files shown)


TannerRogalsky:
We've tested this against the issue from #1691 and it has resolved the issue with no obvious regressions.

__tests__/Auth0Client/getTokenSilently.test.ts
```diff
@@ -1927,13 +1935,13 @@ describe('Auth0Client', () => {
           })
         ]);
 
-        // Both should return the same token due to locking
-        expect(token1).toEqual(token2);
+        // cacheMode: 'off' calls each make their own request, so they get different tokens
```
gyaneshgouraw-okta, line 1938: @yogeshchoudhary147 I feel this is a behavior change worth calling out explicitly in the PR description. May be a new section under behaviour update or any relevant wording that makes it explicitly clear.

Previously, singlePromise keyed only on `clientId::audience::scope`. So N concurrent `getTokenSilently({ cacheMode: 'off' })` calls with the same audience + scope were grouped into a single token request. 

With singlePromise removed, the lock now serialises those calls but each performs its own request, so N concurrent off calls → N token exchanges.
As correctly updated in the test case behaviour.



Additionally minor suggestion, we can explictly callout also in GetTokenSilentlyOptions here 
https://github.com/auth0/auth0-spa-js/blob/fix/cache-mode-off-single-promise/src/global.ts#L552

Update the current `off` value description to

> When off, ignores the cache and always sends a request to Auth0. Concurrent off calls are not deduplicated, each makes its own request.

gyaneshgouraw-okta (dismissed):

gyaneshgouraw-okta (approved):

TannerRogalsky:
Could we cut a release with this change? @yogeshchoudhary147

catnipan:
Could you also cut a new `@auth0/auth0-react` release that bundles this @auth0/auth0-spa-js@2.24.1 (or newer) and bumps the dependency floor from ^2.23.0 accordingly? @yogeshchoudhary147 

the package bundles `@auth0/auth0-spa-js` in so we can't do pnpm override

TannerRogalsky:
@yogeshchoudhary147 Please tell your boss that we think you kick ass.

PR #1733, opened 2026-08-29
fix: widen getTokenSilently overload return types to include undefined

yogeshchoudhary147:
## Summary

The public overloads for `getTokenSilently` declare `Promise<string>` and `Promise<GetTokenSilentlyVerboseResponse>`, but the implementation can return `undefined` in two cases:

- **`cacheMode: 'cache-only'` with no matching cache entry** — pre-existing behaviour; the overloads were already incorrect before IPSIE.
- **IPSIE `session_expiry` ceiling reached** — introduced when the session expiry ceiling feature was added.

The implementation signature already reflects this (`Promise<undefined | string | GetTokenSilentlyVerboseResponse>`), but overload signatures are what callers see — the implementation signature is invisible externally.

This aligns the overloads with the implementation so consumers can write correct null checks without fighting the type system. Additionally, `Fetcher.prepareRequest` now throws `GenericError('missing_access_token')` instead of a `TypeError` when the token factory resolves to `undefined`, giving callers a structured error they can handle.

## Test plan

- [x] Existing unit tests pass (`npm test`)
- [x] TypeScript consumers can correctly narrow `string | undefined` from `getTokenSilently`
- [x] `prepareRequest` rejects with `GenericError` (error: `missing_access_token`) when the token factory returns `undefined`

> Reopened from #1646 after rebasing onto latest `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when silent token retrieval cannot provide an access token.
  * Requests now return a clear `missing_access_token` error instead of proceeding without authorization.

* **Documentation**
  * Updated API examples to safely handle unavailable access tokens and skip requests when necessary.
  * Clarified error handling for cache-only misses and session limits.
  * Public token retrieval results now indicate when no token is available.

### DIFF
--- a/__tests__/fetcher.test.ts
+++ b/__tests__/fetcher.test.ts
@@ -5,7 +5,7 @@
  */
 
 import { beforeEach, describe, expect } from '@jest/globals';
-import { UseDpopNonceError } from '../src/errors';
+import { GenericError, UseDpopNonceError } from '../src/errors';
 import { GetTokenSilentlyVerboseResponse } from '../src/global';
 import {
   Fetcher,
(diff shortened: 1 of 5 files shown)


__tests__/fetcher.test.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import { beforeEach, describe, expect } from '@jest/globals';
-import { UseDpopNonceError } from '../src/errors';
+import { GenericError, UseDpopNonceError } from '../src/errors';
```
highfive251677-wq, line 8:
```suggestion
import { GenericError, UseDpopNonceError } from '../src/errors';
```

cschetan77:
Since `getTokenSilently` public api is changed, can we update the example [calling-an-api.md](https://github.com/auth0/auth0-spa-js/blob/main/examples/calling-an-api.md) to put a guard.

cschetan77 (approved):

PR #1743, opened 2026-09-03
fix: propagate the rotated refresh token to all cache entries on MFA grants

### LINKED ISSUE #1742: MFA completion grant leaves the rotated-away refresh token in other cache entries; next refresh trips reuse detection and revokes the family
### Checklist
- [ ] The issue can be reproduced in the auth0-spa-js sample app (or N/A)
- [x] I have looked into the Readme, Examples, and FAQ and have not found a suitable solution
- [x] I have looked into the documentation and API documentation and have not found suitable answer
- [x] I have searched the issues and have not found a suitable solution or answer
- [x] I have searched the Auth0 Community forums and have not found suitable solution
- [x] I agree to the terms within the Auth0 Code of Conduct

### Description

1. With rotation on, Auth0 invalidates the previous refresh token every time it grants a new one. Presenting the old one afterwards is treated as reuse and revokes the whole family.
2. So whenever the SDK receives a new refresh token, it has to replace every cached copy, regardless of which audience/scope entry made the request. The SDK already does this after a refresh exchange (`_getTokenUsingRefreshToken` → `cacheManager.updateEntry`, #1503 / #1702).
3. After an MFA completion grant it doesn't. `mfa.verify` → `_requestTokenForMfa` → `_requestToken` → `_saveEntryInCache` stores the new refresh token only in the entry for the requested scope/audience. The other entries keep the invalidated token.

Result: the next refresh from any other entry fails (`ferrt` "reused refresh token detected", `tokenCounter: N, latestCounter: N+1`), the family is revoked, and the SDK falls back to iframe auth. If the entry being renewed is MFA-gated by an Action, the iframe fails with "Multifactor authentication required" and the SDK logs the user out mid-flow.

Expected: the MFA grant converges the cache the same way a refresh exchange does. Fix in #1743.

### Reproduction

Consistent. `useRefreshTokens: true`, `useMrrt: true`, rotation on, an Action that calls `api.multifactor.enable` on refresh exchanges requesting a write scope.

1. Log in, then `getTokenSilently()` for a second audience. Both cache entries hold the same refresh token.
2. `getTokenSilently()` for a scope the Action challenges → `MfaRequiredError`; complete it with `mfa.verify(...)`.
3. Inspect the cache: only the new entry has the new refresh token.
4. `getTokenSilently({ cacheMode: 'off' })` for the first entry → `invalid_grant`, tenant log `ferrt`.

### Additional context

Seen through auth0-react 2.24.0. Related: #1507 / #1503, #1684 / #1702.

### auth0-spa-js version
2.24.1

### Framework selection
React (auth0-react)

### Framework version
2.24.0

### Browser testing
Chrome (cache logic is browser-independent)

UTkzhang:
## Problem

With refresh token rotation and `useMrrt: true`, every cache entry shares one refresh token family. When a refresh exchange is refused with `mfa_required` and the app completes it through `mfa.verify(...)`, the completion grant rotates the refresh token like any refresh exchange does, but `_requestTokenForMfa` → `_requestToken` → `_saveEntryInCache` writes the new token only into the entry it just minted. Every other entry keeps the rotated-away token. The next refresh from one of them makes the server log `ferrt` "reused refresh token detected" (`tokenCounter: N, latestCounter: N+1`) and revoke the whole family; the SDK then survives on the iframe fallback, and when the entry being renewed needs a scope the tenant MFA-gates, the iframe fails with "Multifactor authentication required" and the SDK logs the user out.

Reproduction and tenant-log signature: #1742.

**Root cause:** `_getTokenUsingRefreshToken` has converged the cache after a refresh via `cacheManager.updateEntry(...)` since #1503 / #1702. `_requestTokenForMfa` never got the same step.

## Fix

`_requestTokenForMfa` reads the entry the refused exchange borrowed its refresh token from, using the same MRRT-aware `cacheManager.get(...)` lookup `_getTokenUsingRefreshToken` uses, then after `_requestToken` calls `updateEntry(previous.refresh_token, result.refresh_token, clientId, useMrrt)` under the same `!this.onlineAccess` guard. Under MRRT that converges every entry; without MRRT it updates only the entries that held the consumed token.

Notes for review:
- Touches the token cache (the "ask first" area in `CLAUDE.md`), but only repeats the propagation step the refresh path already performs, with the same guards.
- No public API or option changes, so no README / `examples/` update.
- `_requestTokenForPasskey` is left alone: a passkey login starts a new family rather than rotating one an existing entry holds. Happy to extend if you'd rather have it there too.

## Test plan

- [x] `npx jest __tests__/Auth0Client/requestTokenForMfa.test.ts` — two new tests drive the real client through `loginWithRedirect` → MRRT refresh → `mfa_required` refusal → `mfa.verify` → refresh of the untouched entry, and assert which refresh token went out on the wire. Both fail on `main` (the rotated-away token is sent) and pass with the change; the second covers the non-MRRT downscoping shape from #1503.
- [x] `npx jest __tests__/Auth0Client/onlineAccess.test.ts __tests__/mfa` — existing MFA and online-access suites green.
- [x] `npm test` — full unit suite green (51 suites, 1094 tests).
- [x] `npm run build` including `test:es-check`.

Fixes #1742.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved multi-factor authentication token exchanges when refresh-token rotation is enabled.
  - Updated related cached sessions with the newly issued refresh token, helping prevent authentication failures during subsequent token refreshes.
  - Ensured sessions sharing refresh-token credentials continue using the latest valid token after MFA completion.
  - Improved consistency across repeated MFA-related token exchanges when sessions share refresh-token credentials.

### DIFF
--- a/__tests__/Auth0Client/requestTokenForMfa.test.ts
+++ b/__tests__/Auth0Client/requestTokenForMfa.test.ts
@@ -1,0 +1,215 @@
+import { verify } from '../../src/jwt';
+import { MessageChannel } from 'worker_threads';
+import * as utils from '../../src/utils';
+import * as scope from '../../src/scope';
+import { MfaRequiredError } from '../../src/errors';
+
+import { fetchResponse, loginWithRedirectFn, setupFn } from './helpers';
+
+import {
+  TEST_CODE_CHALLENGE,
+  TEST_ID_TOKEN,
+  TEST_REFRESH_TOKEN
+} from '../constants';
+
+jest.mock('es-cookie');
+jest.mock('../../src/jwt');
+jest.mock('../../src/worker/token.worker');
+
+const mockWindow = <any>global;
+const mockFetch = <jest.Mock>mockWindow.fetch;
+const mockVerify = <jest.Mock>verify;
+
+jest
+  .spyOn(utils, 'bufferToBase64UrlEncoded')
+  .mockReturnValue(TEST_CODE_CHALLENGE);
+
+jest.spyOn(utils, 'runPopup');
+
+const setup = setupFn(mockVerify);
+const loginWithRedirect = loginWithRedirectFn(mockWindow, mockFetch);
+
+const API_AUDIENCE = 'https://api.example.com';
+const MY_ACCOUNT_AUDIENCE = 'https://auth0_domain/me/';
+const MFA_TOKEN = 'mfa-token-abc123';
+
+const requestBody = (body: string): Record<string, string> => {
+  try {
+    return JSON.parse(body);
+  } catch {
+    return Object.fromEntries(new URLSearchParams(body).entries());
+  }
+};
+
+const refreshTokenSentInCall = (callNum: number) =>
+  requestBody(mockFetch.mock.calls[callNum][1].body).refresh_token;
+
+// Echoes the requested scope back: the MRRT path rejects a response that does
+// not cover every scope it asked for.
+const mockTokenResponse = (refresh_token: string) =>
+  mockFetch.mockImplementationOnce((_url: string, init: { body: string }) =>
+    fetchResponse(true, {
+      id_token: TEST_ID_TOKEN,
+      access_token: `access-token-for-${refresh_token}`,
+      refresh_token,
+      token_type: 'Bearer',
+      expires_in: 86400,
+      scope: requestBody(init.body).scope
+    })
+  );
+
+const mockMfaRequired = () =>
+  mockFetch.mockResolvedValueOnce(
+    fetchResponse(false, {
+      error: 'mfa_required',
+      error_description: 'Multifactor authentication required',
+      mfa_token: MFA_TOKEN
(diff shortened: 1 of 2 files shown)


yogeshchoudhary147:
Thanks for the contribution, Kevin! The fix looks good. Before we can merge, could you GPG-sign the commits? Our branch protection requires verified commits.

yogeshchoudhary147:
Thanks for the thorough investigation and well-written fix, @UTkzhang — the root cause analysis and test plan are spot on. The CI failures here (Angular Tests, vulnerability check) are pre-existing and unrelated to your change; the fix itself is correct and all unit tests pass locally.

I'm taking this from here and will merge it from a branch on my end so we can get it shipped cleanly. Closing this in favour of a new PR — really appreciate you surfacing this one.

PR #1744, opened 2026-09-04
fix(ci): upgrade Node.js to 24 to fix npm Arborist bug in Angular Tests

yogeshchoudhary147:
## Summary
- Angular Tests started failing due to a known npm Arborist bug (https://github.com/npm/cli/issues/9787) — `ng new`'s internal `npm install` crashes with `Cannot read properties of null (reading 'edgesOut')` on npm 10.x (shipped with Node 22)
- Fixes by upgrading the test workflow to Node 24 (current LTS), which ships with npm 11.x where the Arborist bug is resolved
- Node 22 is in maintenance mode — this also aligns CI with the current LTS lifecycle

## Test plan
- [x] All jobs pass on this PR with Node 24

## Summary by CodeRabbit

* **Chores**
  * Updated the automated testing environment to use Node.js 24.
  * Simplified Angular app setup by restoring dependency installation during project creation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,8 +193,9 @@ jobs:
         with:
           node: ${{ env.NODE_VERSION }}
           cache: ${{ env.CACHE_KEY }}
+          ignore-scripts: false
           install: |
-            npx -p @angular/cli@22.1.6 ng new my-app --defaults=true < /dev/null
+            npx -p @angular/cli ng new my-app --defaults=true < /dev/null
           content: |
             echo -e "${{ env.AUTH0_CONTENT }}" > src/auth0.js;
           import: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,9 +193,9 @@ jobs:
         with:
           node: ${{ env.NODE_VERSION }}
           cache: ${{ env.CACHE_KEY }}
-          ignore-scripts: false
           install: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -194,7 +194,8 @@ jobs:
           node: ${{ env.NODE_VERSION }}
           cache: ${{ env.CACHE_KEY }}
           install: |
-            npx -p @angular/cli ng new my-app --defaults=true < /dev/null
+            npx -p @angular/cli ng new my-app --defaults=true --skip-install < /dev/null
+            npm install --prefix my-app --legacy-peer-deps
```
gyaneshgouraw: Hey @yogeshchoudhary147 wouldn't adding `--legacy-peer-deps` mask dependency resolution issues, which we would normally like to surface.

gyaneshgouraw:
@yogeshchoudhary147 Can we also link the original upstream issue in the PR description for more context?

.github/workflows/test.yml
yogeshchoudhary147: Hey @gyaneshgouraw-okta you are right, this is not the cleanest fix since CI is now validating with the older algorithm while real users would use the strict one. This is a temporary workaround to unblock things until npm fixes the Arborist bug in a newer version, we should revisit this once that happens.

yogeshchoudhary147:
Hey @gyaneshgouraw-okta linked the upstream npm issue in the PR description: https://github.com/npm/cli/issues/9787

PR #1750, opened 2026-09-08
feat(anonymous): add anonymous sessions support

yogeshchoudhary147:
## Summary

- Adds `auth0Client.anonymous` sub-client (`AnonymousSessionApiClient`) with localStorage/memory caching, expiry-aware token refresh, and silent session replacement
- Adds two new config options to `Auth0ClientOptions`: `anonymousSessionsCacheMode` (`'localStorage'` | `'memory'`, defaults to `'localStorage'`) and `createAnonymousSessionOnFailedSilentAuth` (defaults to `false`)
- `checkSession()` will now create an anonymous session when `createAnonymousSessionOnFailedSilentAuth` is enabled, but only when the error is `login_required` (not `consent_required` or `interaction_required` since those mean a user already exists)
- Each `audience` and `scope` combination gets its own cache slot so tokens are never mixed up across resource servers. The same anonymous identity (session token) is shared across all slots
- Re-exports anonymous session types and errors from `@auth0/auth0-auth-js` so consumers do not need to add it as a direct dependency
- Bumps `@auth0/auth0-auth-js` to `^1.14.0`

## Test plan

- [x] All existing unit tests pass (`npm test`)
- [x] `auth0.anonymous.createSession()` creates a session and saves it to localStorage
- [x] `auth0.anonymous.getTokenSilently()` returns the cached token without a network call when the access token is still fresh
- [x] `auth0.anonymous.getTokenSilently()` renews via session token when the access token is expired
- [x] `auth0.anonymous.getTokenSilently({ audience: 'A' })` and `getTokenSilently({ audience: 'B' })` fetch independently with no cross-audience cache hit
- [x] `auth0.anonymous.logout()` clears the stored session
- [x] `createAnonymousSessionOnFailedSilentAuth: true` creates an anonymous session on `checkSession()` for a user who has never logged in
- [x] `loginWithRedirect()` carries the `auth0_anon` cookie automatically

## Summary by CodeRabbit

- **New Features**
  - Added support for anonymous sessions through the Auth0 client.
  - Anonymous sessions can be created or restored after failed silent authentication when enabled.
  - Anonymous session tokens can be stored in memory or local storage.
  - Added configuration options for anonymous session behavior and enterprise connection scenarios.
- **Bug Fixes**
  - Prevented concurrent requests from creating multiple anonymous identities.
  - MFA step-up authentication failures no longer trigger anonymous session creation.
  - Improved handling of anonymous session storage and cleanup errors.


src/anonymous/AnonymousSessionApiClient.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  constructor(
+    private authJsClient: AnonymousSessionClient,
+    clientId: string,
+    cacheMode: 'localStorage' | 'memory' = 'localStorage'
+  ) {
+    const canUseLocalStorage =
+      cacheMode === 'localStorage' &&
+      typeof window !== 'undefined' &&
+      !!window.localStorage;
```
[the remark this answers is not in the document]
yogeshchoudhary147: Rejecting — consistent with the rest of the SDK. The existing `LocalStorageCache` (used for auth tokens when `cacheLocation: 'localstorage'`) has no such guard and has shipped in production for years, meaning Auth0's browser support baseline doesn't include environments where `window.localStorage` throws on property access. Our `makeLocalStore` already wraps every individual operation in try/catch, which covers the modern Safari private-mode failure mode (`setItem` throws, not the property access). Adding a try/catch to the constructor check would be inconsistent with the rest of the SDK.

src/anonymous/AnonymousSessionApiClient.ts
```diff
@@ -0,0 +1,138 @@
...
+    });
+    this.store.set(session);
+    return session;
+  }
+
+  /**
+   * Ends the anonymous session and clears locally stored tokens.
+   */
+  async logout(): Promise<void> {
+    await this.authJsClient.logout();
```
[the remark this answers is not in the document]
yogeshchoudhary147: Rejecting. If `authJsClient.logout()` fails (e.g. network error), the `auth0_anon` cookie is still active on Auth0's domain. Clearing local state unconditionally in a `finally` block would leave the client in an inconsistent state: the next `getTokenSilently()` call would create a brand-new anonymous identity while the old cookie remains live, and the subsequent `loginWithRedirect()` would link the old identity (via the still-active cookie) rather than the new one — a silent identity mismatch. Preserving local state on failure lets the developer retry the logout with the correct session token still in hand. The current sequential behavior (clear local only after a successful remote logout) is intentional.

src/Auth0Client.ts
```diff
@@ -338,6 +349,11 @@ export class Auth0Client {
       clientId: this.options.clientId,
     });
     this.mfa = new MfaApiClient(this.authJsClient.mfa, this);
+    this.anonymous = new AnonymousSessionApiClient(
+      this.authJsClient.anonymous,
+      this.options.clientId,
+      this.options.anonymousSessionsCacheMode ?? 'localStorage'
```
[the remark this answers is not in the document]
yogeshchoudhary147: Rejecting. The design doc explicitly specifies `localStorage` as the default for `anonymousSessionsCacheMode` (see the options table in the EA SDK Requirements doc). Changing it to `memory` would contradict the spec and break the intended default UX — anonymous sessions would not survive page reloads without explicit opt-in configuration.

static/anonymous.html
```diff
@@ -0,0 +1,256 @@
...
+          showResponse('Error [' + (e.code ?? e.name) + ']: ' + e.message);
+        }
+      }
+
+      async function getAccessToken(audience) {
+        if (!anonSession) { showResponse('No session. Run Create Session first.'); return; }
+        showOp('anonymous.getAccessToken({ sessionToken, audience: "' + audience + '" })');
+        try {
+          const client = await getClient();
+          const session = await client.anonymous.getAccessToken({
```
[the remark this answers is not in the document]
yogeshchoudhary147: File removed in f777bc0.

src/global.ts
```diff
@@ -362,6 +362,28 @@ export interface Auth0ClientOptions {
...
+  /**
+   * When `true`, automatically creates (or restores) an anonymous session whenever
+   * `getTokenSilently()` fails because the user is not authenticated.
+   *
+   * The anonymous session is available via `auth0.anonymous.getTokenSilently()` after
+   * the call returns. The original error from `getTokenSilently()` is still thrown.
+   *
+   * @default false
+   */
+  createAnonymousSessionOnFailedSilentAuth?: boolean;
```
cschetan77: I see 2 problems here 
1. 

> When `true`, automatically creates (or restores) an anonymous session whenever `getTokenSilently()` fails because the user is not authenticated.

This is an internal detail of `checkSession()` and is correct, but developer could misinterpret it to something like - 
```js
 try {
    await auth0.getTokenSilently();
  } catch (e) {
    if (e.error === 'login_required') {
      // I read the docs. The option says an anonymous session is automatically
      // created whenever getTokenSilently() fails. So it must be ready now.
      const { accessToken } = await auth0.anonymous.getTokenSilently({
        audience: 'https://api.example.com'
      });
      // ← This makes a fresh network call to mint a new anonymous identity.
      //   The option did nothing. Developer has no idea why.
    }
  }
```

We could reframe it something-
```js
/**
   * When `true`, `checkSession()` will automatically create an anonymous session
   * if no authenticated session exists. After `checkSession()` returns, the
   * anonymous token is available via `auth0.anonymous.getTokenSilently()`.
   *
   * This option has no effect when calling `getTokenSilently()` directly
*/
```

2. 

> "The original error from getTokenSilently() is still thrown."

The `checkSession()` never throw any error and swallows all errors from `getTokenSilently()`. 
This can also be misinterpreted by the developer to do something like
```js
try {
    await auth0.checkSession();
  } catch (e) {
    // I was told the error is still thrown — so I handle it here
    if (e.error === 'login_required') {
      showLoginButton(); // ← this never runs
    }
  }
```

__tests__/anonymous/AnonymousSessionApiClient.test.ts
```diff
@@ -0,0 +1,539 @@
...
+      const result = await client.getTokenSilently({ audience: 'https://api-b.example.com' });
+
+      expect(authJsClient.getAccessToken).toHaveBeenCalledWith({
+        audience: 'https://api-b.example.com',
+        sessionToken: freshSession.sessionToken
+      });
+      expect(result).toBe(newSession);
+    });
+
+    it('reuses sessionToken from localStorage on page reload when fetching a new audience', async () => {
```
cschetan77, line 178: These 2 tests 
1. 'reuses sessionToken from localStorage on page reload when fetching a new audience'
2. 'clears localStorage slots written before page reload on logout'

Both tests seed localStorage with a key format that the production code never writes:
`'@@auth0spajs@@::test_client::anonymous::https://api-a.example.com::'`

The actual key written by `AnonymousSessionCacheManager.getStore()` is:

 `'@@auth0spajs@@::test_client::anonymous::["https://api-a.example.com",""]'`

The tests pass because `getAnySessionToken()` and `removeAll()` use a `startsWith` prefix scan that matches both formats. But this means the tests are not verifying the scenario they claim to test, they're passing against data that can never exist in production.

We should look at using the actual key format in both tests.
Or IMO it's better to avoid hardcoding the key entirely, write the session through the real API first, then instantiate a fresh client to simulate the reload. That way the test stays correct if the key format ever changes. 
```js
// write via real API (page load 1)
  const firstClient = makeClient('localStorage');
  authJsClient.getAccessToken.mockResolvedValueOnce(mockSession({ sessionToken: 'persisted_session_token', ... }));
  await firstClient.getTokenSilently({ audience: 'https://api-a.example.com' });

  // fresh client (page load 2)
  const freshClient = makeClient('localStorage');
  await freshClient.getTokenSilently({ audience: 'https://api-b.example.com' });

  expect(authJsClient.getAccessToken).toHaveBeenCalledWith({
    audience: 'https://api-b.example.com',
    sessionToken: 'persisted_session_token'
  });
```

src/anonymous/AnonymousSessionApiClient.ts
```diff
@@ -0,0 +1,124 @@
...
+   * Creates a new anonymous session and persists the tokens locally.
+   */
+  async createSession(
+    options?: CreateAnonymousSessionOptions
+  ): Promise<AnonymousSession> {
+    const session = await this.authJsClient.createSession(options);
+    this.cache
+      .getStore(
+        (options as AnonymousGetTokenSilentlyOptions)?.audience,
+        (options as AnonymousGetTokenSilentlyOptions)?.scope
```
cschetan77: `options: CreateAnonymousSessionOptions` already has `audience` and `scope` on it. 
Why are we are casting it to `AnonymousGetTokenSilentlyOptions` 
Shouldn't this work ?

```suggestion
        this.cache.getStore(options?.audience, options?.scope).set(session);
```

src/Auth0Client.ts
```diff
@@ -895,9 +913,21 @@ export class Auth0Client {
    *
    * @param options
    */
+  private async _maybeCreateAnonymousSession() {
+    if (this.options.createAnonymousSessionOnFailedSilentAuth) {
+      try {
+        await this.anonymous.getTokenSilently();
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.debug('[auth0-spa-js] Anonymous session creation failed', e);
```
cschetan77, line 929: `_maybeCreateAnonymousSession` calls `getTokenSilently()` with no audience or scope. The resulting token targets the tenant's default audience and is not usable for calling any specific resource server. After checkSession returns, the app still needs to call `auth0.anonymous.getTokenSilently({ audience: '...' })` explicitly, which will make a network call because the no-audience slot and the audience-specific slot are separate cache entries.
The design here is correct and the `checkSession` has no context about what audience to request, but the purpose of the trigger is not documented anywhere, it establishes the anonymous identity (sessionToken) so the subsequent audience-specific call can reuse the same identity without minting a new one. Without this explanation, a developer will be confused by the extra network call after checkSession and may conclude the feature is not working as it should. 

We should clearly explain in the JSDoc for `createAnonymousSessionOnFailedSilentAuth` that after `checkSession` returns, a follow-up call to `auth0.anonymous.getTokenSilently({ audience })` is still needed to get a usable API token.


```js
/**
   * After `checkSession()` returns, the anonymous identity is established — call `auth0.anonymous.getTokenSilently({ audience })`
   * to get an API-scoped token. That call will reuse the identity created here
   * without minting a new one.
*/
```

cschetan77:
PR introduces a significant new public API surface: 
- Two new createAuth0Client options: anonymousSessionsCacheMode, createAnonymousSessionOnFailedSilentAuth
- New auth0.anonymous sub-client with four public methods: createSession(), getTokenSilently(), logout(), getClaims()

can we also look at updating - 
* A section in `README.md` covering the two new options and the auth0.anonymous API
* An `examples/anonymous-sessions.md` with worked examples covering the key flows: `checkSession` with `createAnonymousSessionOnFailedSilentAuth`, explicit `createSession` with metadata, `getTokenSilently` with an audience, and logout

src/global.ts
yogeshchoudhary147: Good catch, updated the doc to make it clear this only fires inside `checkSession()`.

__tests__/anonymous/AnonymousSessionApiClient.test.ts
yogeshchoudhary147: Fixed, the keys now use the correct `["audience","scope"]` JSON format.

src/anonymous/AnonymousSessionApiClient.ts
yogeshchoudhary147: Good catch, removed the cast.

src/Auth0Client.ts
yogeshchoudhary147: Added a comment explaining this. The session token is what we are after here. The app calls `anonymous.getTokenSilently({ audience })` separately and reuses it.

examples/anonymous-sessions.md
```diff
@@ -0,0 +1,93 @@
...
+## Explicit session creation with metadata
+
+Call `auth0.anonymous.createSession()` directly to create an anonymous session with metadata attached to the identity.
+
+```js
+const session = await auth0.anonymous.createSession({
+  metadata: { cart: 'cart-123' }
+});
+
+console.log(session.sessionToken); // opaque JWE — managed by the SDK
```
cschetan77: If we are calling out that this is SDK managed and is completely opaque, we shouldn't indicate or let developer use it any way. 
This line can be removed entirely, or replaced with a `session.accessToken` usage to show the developer what they actually do with the result.

examples/anonymous-sessions.md
```diff
@@ -0,0 +1,93 @@
...
+});
+```
+
+If you use `new Auth0Client()` directly, call `checkSession()` yourself after construction.
+
+This does not affect standalone `getTokenSilently()` calls on `auth0` itself. If you need to attach metadata to the anonymous identity, use explicit session creation instead.
+
+## Explicit session creation with metadata
+
+Call `auth0.anonymous.createSession()` directly to create an anonymous session with metadata attached to the identity.
```
cschetan77, line 37: We should also callout that `metadata` can only be set at creation time. Developers will naturally try to update it later and either get a silent failure or an `invalid_request` error they won't understand.

examples/anonymous-sessions.md
```diff
@@ -0,0 +1,93 @@
...
+});
+
+const { accessToken: tokenB } = await auth0.anonymous.getTokenSilently({
+  audience: 'https://api-b.example.com'
+});
+```
+
+## Ending the session
+
+Call `auth0.anonymous.logout()` to end the anonymous session and clear all locally stored tokens.
```
cschetan77, line 79: The shows `logout()` in isolation with no context of when to call it. The expected developer flow is: call `loginWithRedirect()` first (the cookie links the anonymous identity during login), then after login completes call `auth0.anonymous.logout()` to clean up. Without this, a developer might call logout() before loginWithRedirect() and destroy the identity link they needed. 
We should call out this.

examples/anonymous-sessions.md
```diff
@@ -0,0 +1,93 @@
+# Anonymous Sessions
+
+Anonymous sessions let you assign a persistent identity to a visitor before they log in. The visitor gets an access token tied to an opaque anonymous identity. When they eventually log in, the anonymous identity can be linked to their real account.
```
cschetan77: This implies linking is automatic and seamless. But Auth0 detects the anonymous session during login and surfaces it as `event.anonymous_session` in Actions,  but nothing happens with it unless the developer writes a Post-Login Action.
We should call out this as well.

examples/anonymous-sessions.md
yogeshchoudhary147: Removed it entirely.

examples/anonymous-sessions.md
yogeshchoudhary147: Added a note. Also called out that calling `createSession()` again creates a new identity, not an update.

examples/anonymous-sessions.md
yogeshchoudhary147: Added a note before the snippet about calling `loginWithRedirect()` before `logout()`.

examples/anonymous-sessions.md
yogeshchoudhary147: Rewrote the intro to make it clear nothing is linked automatically and a Post-Login Action is needed.

cschetan77 (dismissed):

examples/anonymous-sessions.md
```diff
@@ -0,0 +1,95 @@
...
+- [Automatic creation via checkSession](#automatic-creation-via-checksession)
+- [Explicit session creation with metadata](#explicit-session-creation-with-metadata)
+- [Getting an access token](#getting-an-access-token)
+- [Multiple audiences](#multiple-audiences)
+- [Ending the session](#ending-the-session)
+- [Storage modes](#storage-modes)
+
+## Automatic creation via checkSession
+
+Set `createAnonymousSessionOnFailedSilentAuth: true` to have the SDK automatically create an anonymous session when `checkSession()` finds no authenticated user. `createAuth0Client()` calls `checkSession()` internally, so the anonymous session is ready as soon as the client is initialized.
```
gyaneshgouraw, line 16: I feel this line is internal implementation detail, should we highlight this ?
> `createAuth0Client()` calls `checkSession()` internally, so the anonymous session is ready as soon as the client is initialized.

README.md
```diff
@@ -165,6 +165,31 @@ Pass these **per-call** rather than in the `authorizationParams` at client const
 
 **For production:** pass the variant decision from a feature-flag tool (e.g. LaunchDarkly) that has already decided which variant the user should see.
 
+### Anonymous Sessions
```
gyaneshgouraw, line 168: I feel we should rethink adding feature details in readme. I understand  couple of features are already added, and this also follows the same pattern.
But adding feature details here will eventually lead to bloated readme. 

We can do with 1 sentence entry in readme and redirect to corresponding example.

src/Auth0Client.ts
```diff
@@ -912,7 +946,15 @@ export class Auth0Client {
 
     try {
       await this.getTokenSilently(options);
-    } catch (_) { }
+    } catch (e) {
+      if (
+        e instanceof GenericError &&
+        e.error === 'login_required' &&
+        e.error_description !== MFA_STEP_UP_ERROR_DESCRIPTION
```
gyaneshgouraw, line 956: Hey @yogeshchoudhary147 we already have a helper class to capture this inline MFA detection
```  
e.error === 'login_required' &&
e.error_description !== MFA_STEP_UP_ERROR_DESCRIPTION
```
called `_isIframeMfaError`, please review this, so we have single source to capture these errors.

cschetan77 (approved):

src/Auth0Client.ts
yogeshchoudhary147: Agree on single source of truth. Skipped `_isIframeMfaError` because the name is iframe-specific and this callsite has no iframe context. Renaming it to `_isMfaStepUpError` would make it usable in both places. Happy to do that as a follow-up.

README.md
yogeshchoudhary147: Fair point. Let's do a README cleanup as a separate exercise so all sections are treated consistently.

examples/anonymous-sessions.md
yogeshchoudhary147: This is in the examples doc where explaining the mechanism is useful. It is why the `new Auth0Client()` note below makes sense. Keeping it.

src/Auth0Client.ts
gyaneshgouraw: I see `checkSession() → getTokenSilently()`.  Where getTokenSilently() will run into iframe path.
And then only we will run into this error, so this callsite touches iframe in my understanding.

yogeshchoudhary147: Makes sense. Will fix this in a follow-up PR.

yogeshchoudhary147: We cannot directly replace with `!_isIframeMfaError(e)` because we still need the `e.error === 'login_required'` guard. Without it, `!_isIframeMfaError(e)` would also be true for `consent_required` and `interaction_required` errors, which would incorrectly trigger anonymous session creation for users who already exist.

